PR #4, opened 2023-10-13
Update action tags to prefer actions using node20 

erikburt:
These 3 action references use `node16` but their main branch has been updated to use `node20`.

```
golangci/golangci-lint-action@3a919529898de77ec3da873e3063ca4b10e7f5cc -> A:Run golangci-lint -> node16'

actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe -> A:Setup Go environment -> node16'

actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 -> A:Checkout -> node16'
```

[RE-1955]: https://smartcontract-it.atlassian.net/browse/RE-1955

### DIFF
--- a/.github/workflows/pull-request-main.yml
+++ b/.github/workflows/pull-request-main.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo (needed to reference local action)
-        uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # v4.1.0
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
       - name: ci-lint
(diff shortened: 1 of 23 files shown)


momentmaker (approved):

erikburt:
I need to double check with the renovate bot about how these references are updated.

The problem is, I'm tagging to specific SHAs, but they are not actually included in a tagged release. It seems like we specify a version and the renovate bot will update the SHA tag as releases are made for the version specified.

With my current changes it may actually break that functionality. Going to mark as draft for now.

erikburt:
> I need to double check with the renovate bot about how these references are updated.

Renovate bot hasn't been added to this repo since it was made public. I've opened a request to get that added.

erikburt:
Going to split this up into multiple PRs so the main branch changeset is more atomic.

PR #25, opened 2023-11-04
pin version for push-gha-metrics-action

momentmaker:


momentmaker:
I'm using `actionlint` which lints github actions. It should be good right?

erikburt:
> I'm using `actionlint` which lints github actions. It should be good right?

I'm not really sure, it seems like a widely used tool so very unlikely it actually breaks anything.

momentmaker:
`if break then destroy actionlint`

PR #53, opened 2023-11-27
add manifest action

ttata-cll:
(opening from a fork because I don't have write access to this repo)

Our team needs a workflow that will automatically build all nix packages in a given repo. We want to publish this as a re-usable action so we can share it across repos. https://smartcontract-it.atlassian.net/browse/INT2-413


momentmaker:
nice! could you please add a changeset to this new action?

ttata-cll:
> nice! could you please add a changeset to this new action?

sure! Are there instructions on how to do that? Sorry, I've never used that tool before

momentmaker:
yeah you can do `pnpm i` first and then `pnpm changeset` at the root and check the action you just added :)

actions/manifest-build/action.yml
```diff
@@ -0,0 +1,75 @@
...
+    required: false
+secrets:
+  github-token:
+    description: "GitHub auth token for private repos (if needed)"
+    required: false
+    default: ""
+  cache-token:
+      description: "Cachix auth token for private cache"
+      required: false
+      default: ""
```
ro-tex, line 29: Should we add an S3 option as well or is it too early?

ttata-cll: I think we should hold off and add that once we can test it e2e

momentmaker (approved):

PR #64, opened 2023-12-02
feature/signed commits

HenryNguyen5:
- Add signed-commits
- Auto commit build artifact in PRs

## Testing

Tested a lot of this functionality manually with this repo: https://github.com/smartcontractkit-test/monorepo-test

#### Initial testing / setup
1. Used https://github.com/smartcontractkit/releng-ts-app as a base for that repository
2. To bypass needing to setup GATI
	1. Created an FGAT with `contents:write` `pull-requests:write` `metadata:read` for the `monorepo-test` repository
	2. Added that as a secret to the repository
3. Created a branch in this repo (https://github.com/smartcontractkit/.github/tree/test/cicd-changesets-signed-commits) which updates the `cicd-changesets` action to point to `signed-commits` rather than `changesets/action` ([diff with main](https://github.com/smartcontractkit/.github/compare/main...test/cicd-changesets-signed-commits))
4. Pointed the `push-main` workflow `monorepo-test` to the modified `cicd-changesets` (signed commits)
    - Successful run for `push-main`: https://github.com/smartcontractkit-test/monorepo-test/actions/runs/7189296320

#### Subsequent testing
1. Created PR to version bump both packages: https://github.com/smartcontractkit-test/monorepo-test/pull/1
2. Merged PR to main to trigger a "Version packages" PR - https://github.com/smartcontractkit-test/monorepo-test/pull/2
    * https://github.com/smartcontractkit-test/monorepo-test/actions/runs/7189517886
	* https://github.com/smartcontractkit-test/monorepo-test/pull/2 (note this was opened by me because we are using my FGAT)
3. Merged "Version packages" PR
    * Kicked off `push-main` workflow: https://github.com/smartcontractkit-test/monorepo-test/actions/runs/7189535843
    * Which kicked off two `push-tag-release` workflows:
        * https://github.com/smartcontractkit-test/monorepo-test/actions/runs/7189540583
        * https://github.com/smartcontractkit-test/monorepo-test/actions/runs/7189540587
4. Two tags/releases created which are "verified"
    * https://github.com/smartcontractkit-test/monorepo-test/releases/tag/releng-ts-app-example%400.2.0
    * https://github.com/smartcontractkit-test/monorepo-test/releases/tag/releng-ts-app-cli%400.2.0


.github/workflows/pull-request-main.yml
```diff
@@ -39,6 +39,43 @@ jobs:
           gc-basic-auth: ${{ secrets.GRAFANA_CLOUD_BASIC_AUTH }}
           gc-host: ${{ secrets.GRAFANA_CLOUD_HOST }}
 
+  cd-signed-commits:
```
erikburt, line 42: Should this depend on `ci-signed-commits` so we don't "deploy" failing code?

actions/signed-commits/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
```
erikburt, line 56: Ideally all these action references are at least pointing to the latest major version.

actions/signed-commits/action.yml
```diff
@@ -0,0 +1,43 @@
+name: Changesets
+description: A GitHub action to automate releases with Changesets
+runs:
+  using: "node16"
```
erikburt, line 4: Can we use `node20`?

actions/signed-commits/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "@types/semver": "^7.5.6",
+    "@types/unist": "^2.0.10",
+    "fixturez": "^1.1.0",
+    "nock": "^13.3.8",
+    "node-fetch": "^2.0.0",
+    "prettier": "^2.0.0",
+    "typescript": "^5.3.2"
+  },
+  "engines": {
+    "node": ">= 16"
```
erikburt, line 25: node 20?

actions/signed-commits/src/git/github-git/repo-status.ts
```diff
@@ -0,0 +1,133 @@
...
+      indexStatus: code.charAt(0),
+      workingTreeStatus: code.charAt(1),
+    };
+  }
+
+  /**
+   *  Expects the output of `git status --porcelain=v1` as input
+   *
+   * @param output
+   * @returns
```
erikburt, line 91: Finish or delete

actions/signed-commits/src/readChangesetState.ts
erikburt, line 1: This filename is using camelCase but the rest are using kebab-case

actions/signed-commits/tsconfig.spec.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "../../dist/out-tsc",
+    "module": "commonjs",
```
momentmaker, line 5: this is redundant no as it inherit from the file above?

actions/signed-commits/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@smartcontractkit/changesets-action-signed-commits",
+  "version": "1.4.5",
```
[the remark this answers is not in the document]
erikburt: Downgrading to `0.1.0` - the changesets release should bump it to 1.0

.github/workflows/pull-request-main.yml
HenryNguyen5: that's an option, it would serialize the two jobs though and make the feedback loop slower. The building of the action itself can surface errors that need to be fixed. I think it's better that instead we add both of these checks as required checks in order to merge to trunk

erikburt: > that's an option, it would serialize the two jobs though and make the feedback loop slower.

I don't think it would slow the feedback loop enough to cause many issues. My main problem with it is the automation of commits for code that hasn't passed CI.

Why add that to the commit history?

HenryNguyen5: IMO we're at a edge case b/c usually you never commit artifacts to a repository in the first place. I think there's value in trying to build anyway, we often run docker builds as a CI check. I see your point about commit history pollution, but if it becomes a large problem then we could squash the PR.

I think its because i see it as a CI check of its own which is why IDM leaving it in the way it is. I'd have a problem with it if it was auto deploying too, but we're not doing that, nor are we affecting other people outside of this PR scope.

I also think we should always have CI as fast as possible, shorter loops have an outsized impacts on dev workflow (1 min feedback vs 2 min is actually a huge difference)

erikburt: > I think its because i see it as a CI check of its own which is why IDM leaving it in the way it is. 

I'm not against building in CI, I'm against automatically committing compiled code to a PR. In fact, we're already building as part of a pull request CI.

- https://github.com/smartcontractkit/.github/blob/edf5b41acd7ae24807396c43c967800415f14cd1/.github/workflows/pull-request-main.yml#L65
- https://github.com/smartcontractkit/.github/blob/edf5b41acd7ae24807396c43c967800415f14cd1/actions/cicd-build-publish-artifacts-ts/action.yml#L109
- https://github.com/smartcontractkit/.github/blob/edf5b41acd7ae24807396c43c967800415f14cd1/package.json#L10C1-L10C58

HenryNguyen5: good point, lets serialize it then

HenryNguyen5: Resolving this b/c we already do a build inside the tests themselves, oops :see_no_evil:

erikburt: Discussed offline. 

Will follow-up with an updated workflow for building + committing outputs for all typescript actions with more detailed messages.

erikburt: See 70e65bbc3f1926119ed93a2f4962cf2d7198cc61 and #78.

erikburt (approved):

PR #98, opened 2023-12-21
run manifest build in parallel

tt-cll:
We want to allow the manifest builder to evaluate all flakes in parallel via a check.


momentmaker:
@tt-cll could you please add a changeset to this?

PR #80, opened 2023-12-14
partial rollout - cicd-changesets using signed-commits

erikburt:
## This PR branch should not be deleted until repos that reference this branch are updated to reference main

Switching from https://github.com/changesets/action/ to `signed-commits` introduced in #64.

This PR branch will be used as a partial roll-out, applying this change to repositories.

Partial roll-outs so far:
- https://github.com/smartcontractkit/github-app-token-issuer/commit/e5d9e5ef0001147dbbdf4edeada98dfa123f0b49
    - Undo: https://github.com/smartcontractkit/github-app-token-issuer/pull/66
- [smartcontractkit/.github@`16d8dab`](https://github.com/smartcontractkit/.github/commit/16d8dab56873e35e50eeb33ed988df09c7cbbb6c)
    - Undo: #122
- https://github.com/smartcontractkit/mercury-pipeline/commit/e4a38d4b7794598f80ba17d6d5ba737029c14d09
    - Undo: https://github.com/smartcontractkit/mercury-pipeline/pull/683
- https://github.com/smartcontractkit/gauntlet-evm/commit/b3c651a1f7d8b12c13e0686c245cbb955da6a9a0
    - Undo: https://github.com/smartcontractkit/gauntlet-evm/pull/572

PR #120, opened 2024-01-09
feat: cicd-changesets using signed-commits

erikburt:
Full roll-out for `cicd-changesets` with `signed-commits`. This PR uses the new `signed-commits` action which is a drop-in replacement for the `changesets/action`, but offers signed commits and therefore "signed" tags.

#80 is another PR which has been referenced in other repos during the partial roll-out. This partial roll-out confirmed compatibility of the update. As a follow-up, those partially rolled-out repos will need to be updated to reference the official release. The tracking of updating those references will be done in that PR. 

---

https://smartcontract-it.atlassian.net/browse/RE-1901

Related to #80.

### DIFF
--- a/.changeset/soft-radios-breathe.md
+++ b/.changeset/soft-radios-breathe.md
@@ -1,0 +1,5 @@
+---
+"cicd-changesets": minor
+---
+
+Use signed-commits action, drop-in replacement for changesets/action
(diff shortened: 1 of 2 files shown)


momentmaker (approved):
:nice:

PR #137, opened 2024-01-17
config: renovate - skip dev dependency patch updates

erikburt:
To avoid a lot of spam from renovate, we should skip patch updates for dev dependencies.


momentmaker (approved):
:nice:

PR #139, opened 2024-01-18
feat: setup-gap support for k8s

erikburt:
Action updates to support proxying `kubectl`. 

This action adds a couple steps to support `kubectl`. Some things to note:
1. `kubectl` requires an HTTPS connection to the desired endpoint
2. `kubectl` validates the server's certificate against the CA configured in the kube config
3. When creating a kube config entry using `aws eks update-kubeconfig` it will populate a remote endpoint (not publicly accessible), and the `CA` information

So to setup GAP for K8s we have to:
1. Generate an ephemeral CA key/cert which will be
    * Used to sign the generated server's key
    * Used as the CA file within the kube config
2. Generate a server key and CSR, then use the generate CA key to sign server's CSR, producing a certificate
    * This will be used by the local docker container to satisfy kubectl's requirements. All external connections are secured through TLS with actual certs, and further authentication/authorization through AWS
4. Modify the kube config to
    * Point to `localhost` (as requests will be sent to the `aws-sigv4-proxy`
    * Use the CA generated previously
    * Unset the CA information from `aws eks update-kubeconfig`
5. Pull the custom `aws-sigv4-proxy` image which supports TLS (this will be unnecessary if/when https://github.com/awslabs/aws-sigv4-proxy/pull/179 is merged + released)
6. Run the docker container with the above generated/signed server key + cert.

---

https://smartcontract-it.atlassian.net/browse/RE-2134


actions/setup-gap/action.yml
```diff
@@ -76,31 +110,126 @@ runs:
...
         AWS_REGION: us-east-1
       with:
         registry-type: public
 
-    - name: Run aws sig4 proxy container
+    - name: Run aws sig4 proxy container (public ecr only)
+      if: inputs.use-private-ecr-registry != 'true'
       shell: bash
       env:
         AWS_ACCESS_KEY_ID: ${{ env.AWS_ACCESS_KEY_ID }}
```
momentmaker, line 182: where are these getting initially set?

erikburt: The top of the action assumes an aws role.

momentmaker:
The `setup-gap` works great it seems but the test is failing here?
https://github.com/smartcontractkit/releng-test/actions/runs/7633042156/job/20794469999#step:3:1

erikburt:
> The `setup-gap` works great it seems but the test is failing here? https://github.com/smartcontractkit/releng-test/actions/runs/7633042156/job/20794469999#step:3:1

Yeah the role is no longer apart of the k8s cluster’s AWS auth config. So it cannot actually make any calls. The successful run of that workflow is from friday when the aws role had the proper permissions

momentmaker (approved):

actions/setup-gap/action.yml
```diff
@@ -76,31 +110,126 @@ runs:
...
+        fi
+
+    - name: Generate and Sign Server Certificate (K8s only)
+      if: inputs.use-k8s == 'true'
+      shell: bash
+      run: |
+        mkdir -p /tmp/setup-gap
+
+        echo "::debug::Generating server key and certificate signing request (CSR)"
+        openssl genrsa -out /tmp/setup-gap/server.key 2048
```
erikburt, line 136: use ECDSA

actions/setup-gap/action.yml
```diff
@@ -76,31 +110,126 @@ runs:
...
+      run: |
+        mkdir -p /tmp/setup-gap
+
+        if [ -n "${{ inputs.tls-ca-private-key }}" ] && [ -n "${{ inputs.tls-ca-cert }}" ]; then
+          echo "::debug::Using provided CA key+cert. Writing them to /tmp/setup-gap/ca.key and /tmp/setup-gap/ca.crt"
+          echo "${{ inputs.tls-ca-private-key }}" > /tmp/setup-gap/ca.key
+          echo "${{ inputs.tls-ca-cert }}" > /tmp/setup-gap/ca.crt
+        else
+          echo "::debug::Generating new CA key+cert. Writing them to /tmp/setup-gap/ca.key and /tmp/setup-gap/ca.crt"
+          openssl genrsa -out /tmp/setup-gap/ca.key 4096
```
erikburt, line 125: use ECDSA

erikburt: 256

actions/setup-gap/action.yml
erikburt: 256

erikburt:
Leave comments explaining what we're doing and how this is only TLS for a localhost connection.

actions/setup-gap/action.yml
```diff
@@ -79,31 +106,128 @@ runs:
...
+        AWS_SECRET_ACCESS_KEY: ${{ env.AWS_SECRET_ACCESS_KEY }}
+        AWS_SESSION_TOKEN: ${{ env.AWS_SESSION_TOKEN }}
+      run: |
+        docker run --rm -d \
+          -e AWS_ACCESS_KEY_ID \
+          -e AWS_SECRET_ACCESS_KEY \
+          -e AWS_SESSION_TOKEN \
+          -p ${{ inputs.proxy-port }}:${{ inputs.proxy-port }} \
+          -v /tmp/setup-gap:/tls \
+          ${{ inputs.ecr-private-registry }}.dkr.ecr.${{ inputs.ecr-private-aws-region || inputs.aws-region }}.amazonaws.com/aws-sigv4-proxy:${{ inputs.ecr-private-image-tag }} -v \
```
chainchad, line 222:
```suggestion
          ${{ inputs.ecr-private-registry }}.dkr.ecr.${{ inputs.ecr-private-aws-region || inputs.aws-region }}.amazonaws.com/aws-sigv4-proxy:${{ inputs.ecr-private-image-tag }} \
```

Does the `-v` flag here increase verbosity on the sigv4 proxy? If so, I don't think we want that, right?

actions/setup-gap/action.yml
```diff
@@ -79,31 +106,128 @@ runs:
...
+            --name execute-api \
+            --region ${{ inputs.aws-region }} \
+            --host "${{ inputs.api-gateway-host }}" \
+            --log-failed-requests --log-signing-process \
```
chainchad, line 226:
```suggestion
            --log-failed-requests \
```

We definitely don't want to log signing (could dump creds to output).

actions/setup-gap/action.yml
```diff
@@ -79,31 +106,128 @@ runs:
...
+            --port :${{ inputs.proxy-port }} \
```
chainchad, line 227: Ignore if so, but is this value to the --port flag correct?

actions/setup-gap/action.yml
chainchad: Also, we should confirm that `log-failed-requests` does not log anything sensitive. If it does, we should remove it here and the other spot where it's used.

chainchad (changes requested):

actions/setup-gap/action.yml
erikburt: I'm not sure how that got there to be honest. Removing.

actions/setup-gap/action.yml
erikburt: Sounds good - removing `--log-signing-process`.

`--log-failed-requests` is also included in the original `setup-gap` ([link](https://github.com/smartcontractkit/.github/blob/8bfaa70ae578c4a494eeaf7985ae2e379be3665e/actions/setup-gap/action.yml#L105)), and appears to be fine see: [source](https://github.com/awslabs/aws-sigv4-proxy/blob/bebf2d5a8f03013c091ecd1dfbe2bab80bafb383/handler/proxy_client.go#L242-L252).

actions/setup-gap/action.yml
erikburt: IIRC the problem is when using the tls client it requires the `:`. But otherwise, I atleast know that it definitely works with the `:` prefix.

actions/setup-gap/action.yml
erikburt: d9224ca426a84e53b45e1a6d92b042e067eb55fb

actions/setup-gap/action.yml
erikburt: d9224ca426a84e53b45e1a6d92b042e067eb55fb

chainchad (approved):

PR #153, opened 2024-02-01
feat: support grafana org/tenant id input for push-gha-metrics-action v2.1.0

erikburt:
Updated all actions which reference `push-gha-metrics-action` to accept a new *optional* `gc-org-id` input, and pass that to the action which has been updated for support with grafana self-hosted.

Also tried to generify references to grafana *cloud* to just grafana
- Kept inputs prefixed with `gc-` for backwards compatibility as there are many references to these shared actions with `@main`.

### DIFF
--- a/libs/nx-chainlink/src/generators/create-gh-action/files/composite/action.yml.template
+++ b/libs/nx-chainlink/src/generators/create-gh-action/files/composite/action.yml.template
@@ -11,16 +11,19 @@ inputs:
     description: "number of commits to fetch"
     required: false
     default: "0"
-  # grafana cloud inputs
+  # grafana inputs
   metrics-job-name:
     description: "grafana metrics job name"
     required: false
     default: <%= name %>
   gc-host:
-    description: "grafana cloud hostname"
+    description: "grafana hostname"
     required: false
   gc-basic-auth:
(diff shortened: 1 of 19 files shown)


libs/nx-chainlink/src/generators/create-gh-action/files/composite/action.yml.template
```diff
@@ -39,7 +39,7 @@ runs:
     - name: Collect metrics
       if: always()
       id: collect-gha-metrics
-      uses: smartcontractkit/push-gha-metrics-action@031fd89b33b3e7722b0c4d2593ece1861d9f503f
+      uses: smartcontractkit/push-gha-metrics-action@0281b09807758be1dcc41651e44e62b353808c47 # v2.1.0
       with:
         basic-auth: ${{ inputs.gc-basic-auth }}
         hostname: ${{ inputs.gc-host }}
```
tateexon, line 45: Missing the org id

erikburt: Fixed in: d0134afbf8a192da1ace4d1f2fcd2a45183b9695

tateexon (approved):

PR #209, opened 2024-03-06
add llm-pr-writer action

momentmaker:
This adds a new action: `llm-pr-writer`.

### DIFF
--- a/actions/llm-pr-writer/action.yml
+++ b/actions/llm-pr-writer/action.yml
@@ -75,8 +75,10 @@ runs:
         )
 
         # replace lines starting with ' -' (1space) with '  -' (2spaces)
-        response_content=$(echo $prompt_response | jq -r '.choices[0].message.content')
+        response_content=$(echo "$prompt_response" | jq -r '.choices[0].message.content')
         response_content=$(echo "$response_content" | sed -e 's/^ -/  -/g')
+
+        echo $response_content
 
         {
           echo 'response<<EOF'
--- a/actions/llm-pr-writer/action.yml
+++ b/actions/llm-pr-writer/action.yml
@@ -68,8 +68,9 @@ runs:
       shell: bash
       run: |
         diff_output=$(cat diff_output.txt)
+        openai_model=${{ inputs.openai-model }}
         prompt_response=$(echo '{
-          "model": ${{ inputs.openai-model }},
+          "model": "'$openai_model'",
           "temperature": 0.7,
           "messages": [
             {
--- a/actions/llm-pr-writer/action.yml
+++ b/actions/llm-pr-writer/action.yml
@@ -17,14 +17,12 @@ inputs:
   gh-token:
     description: "github token to make a comment with"
     required: true
-    # will require either pull_request: write or issues: write permission
+    # will require at least three permissions:
+    # contents: read
+    # pull-requests: write or issues: write
+    # repository-projects: read
     # https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#create-an-issue-comment
-  exclude-paths:
-    description: "paths to exclude from the analysis"
-    required: false
-    default: ""
-    # Use semicolons (;) to separate multiple paths.
-    # Example: 'poetry.lock;artifacts/*'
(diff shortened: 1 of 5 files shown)


actions/llm-pr-writer/action.yml
```diff
@@ -0,0 +1,138 @@
...
+inputs:
+  # general inputs
+  checkout-repo:
+    description: "enable git checkout repo"
+    required: false
+    default: "true"
+  checkout-repo-fetch-depth:
+    description: "number of commits to fetch"
+    required: false
+    default: "0"
```
erikburt, line 20: In what scenario would people want to checkout vs not checkout?

actions/llm-pr-writer/action.yml
```diff
@@ -0,0 +1,138 @@
...
+  gh-token:
+    description: "github token to make a comment with"
```
erikburt, line 22: Might be worth calling out that it requires `pull_requests: write` or `issues: write`
- https://docs.github.com/en/rest/issues/comments?apiVersion=2022-11-28#create-an-issue-comment

actions/llm-pr-writer/action.yml
```diff
@@ -0,0 +1,138 @@
...
+    required: true
+  exclude_paths:
+    description: "paths to exclude from the analysis"
```
erikburt, line 25: What format are these paths in?

actions/llm-pr-writer/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "llm-pr-writer",
+  "version": "0.0.0",
+  "description": "This GitHub workflow harnesses the capabilities of Large Language Models (LLMs) to automate the creation and enhancement of pull request (PR) descriptions. Designed to streamline the development process, it significantly reduces the time and effort traditionally required to draft comprehensive PR descriptions, particularly for extensive changes. By meticulously analyzing the modifications within a PR, the workflow generates precise, context-rich summaries that facilitate improved collaboration and review efficiency.",
```
erikburt, line 4: nit: this seems unnecessary as it's in the README

actions/llm-pr-writer/action.yml
momentmaker: good point

actions/llm-pr-writer/action.yml
momentmaker: added comment with usage and example.

actions/llm-pr-writer/action.yml
```diff
@@ -0,0 +1,146 @@
...
+        BRANCH: ${{ inputs.checkout-repo-fetch-depth }}
+        PR_NUMBER:
+          ${{ github.event.issue.number || github.event.pull_request.number ||
+          '' }}
+      run: |
+        # base_ref=$(gh pr view --json baseRefName | jq -r '.baseRefName')
+        # exclude_filter=$(cat $GITHUB_OUTPUT | grep exclude_filter | cut -d'=' -f2)
+        # diff_command="git diff origin/$base_ref...HEAD -- $exclude_filter"
+        # echo "$diff_command"
+        # diff_output=$(eval $diff_command)
```
kalanyuz, line 77: I was using `gh pr diff` before. However they currently don't support file exclusion filters and this sometimes resulted in build artifacts being sent as a part of the prompt, causing it to cross the token limit threshold.
We had to go with the elaborate ways of computing the branch split to head and resort to `git diff` to get this functionality.

momentmaker: ahh the reason I commented it out because of this error here:
https://github.com/smartcontractkit/falcon_llm/actions/runs/8180659211/job/22369114574?pr=6#step:2:168

it's about finding the ref for some reason yet it works with the `gh` way.

kalanyuz: so it seems like why that error appeared is due to this line being removed:
```
gh pr checkout ${{ github.event.issue.number || github.event.pull_request.number || '' }}
```
As some how gh CLI needs to check out the PR again (even after running action/checkout) to have other metadata generated for it.

kalanyuz (approved):

erikburt (approved):

PR #236, opened 2024-03-19
add grafana credential check

momentmaker:
This adds another step `check-grafana-credentials` before collect metrics to ensure all the credentials are there before calling the collect metrics step.


momentmaker:
@erikburt @chainchad what do you guys think about this check?

or should we add it here?
https://github.com/smartcontractkit/push-gha-metrics-action/blob/main/action.yml

erikburt:
>  @erikburt @chainchad what do you guys think about this check?

I think if we have to add this to every action then there's a lot of duplication. I guess ideally we update the action's source (the TS not the action.yml) to warn if the required parameters don't exist, instead of trying to make a request and then logging an error?

momentmaker:
cool make sense :)

PR #278, opened 2024-04-08
fix: cicd-changesets upsert pr draft mode

momentmaker:
This fixes the issue where updating an existing PR does not include the draft mode option.

### DIFF
--- a/actions/cicd-changesets/action.yml
+++ b/actions/cicd-changesets/action.yml
@@ -150,6 +150,16 @@ runs:
       run: |
         echo "published: ${{ steps.changesets.outputs.published }}"
         echo "publishedPackages: ${{ steps.changesets.outputs.publishedPackages }}"
+        echo "hasChangesets: ${{ steps.changesets.outputs.hasChangesets }}"
+        echo "pullRequestNumber: ${{ steps.changesets.outputs.pullRequestNumber }}"
+
+    - name: Set PR to draft
+      shell: bash
+      if:
+        ${{ inputs.pr-draft == 'true' &&
+        steps.changesets.outputs.pullRequestNumber != '' }}
+      run: |
+        gh pr ready ${{ steps.changesets.outputs.pullRequestNumber }} --undo
 
     - name: Collect metrics
       if: always()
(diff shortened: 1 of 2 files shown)


actions/cicd-changesets/action.yml
```diff
@@ -150,6 +150,18 @@ runs:
       run: |
         echo "published: ${{ steps.changesets.outputs.published }}"
         echo "publishedPackages: ${{ steps.changesets.outputs.publishedPackages }}"
+        echo "hasChangesets: ${{ steps.changesets.outputs.hasChangesets }}"
+        echo "pullRequestNumber: ${{ steps.changesets.outputs.pullRequestNumber }}"
+
+    - name: Set PR to draft
```
erikburt, line 156: Can we and are we better off doing this in the `signed-commits` action itself?

https://github.com/smartcontractkit/.github/blob/3c553cf15578c321b73edbe74265e4bde2cb6445/actions/signed-commits/src/run.ts#L454-L462

momentmaker: yeah that would be great but it turns out there are only two states: `open` and `closed`

the other solution is do it through gql:
https://github.com/orgs/community/discussions/45174

the regular api doesn't support this:
https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#update-a-pull-request

erikburt: I just updated a test draft PR with:

```
  gh api \
        --method PATCH \
        -H "Accept: application/vnd.github+json" \
        -H "X-GitHub-Api-Version: 2022-11-28" \
        /repos/smartcontractkit/<repo>/pulls/<pull> \
        -f title='updated title'
```

It did not update the state. Seems like we can include this:

```
 state: "open", 
```

if `prDraft === false `

erikburt: Or something like 

```js
const requestObject = { 
   pull_number: pullRequest.number, 
   title: finalPrTitle, 
   body: prBody, 
   state: "open", 
   ...github.context.repo, 
 }

if (prDraft) delete requestObject.state

 await octokit.rest.pulls.update(requestObject);
```

erikburt: Didn't  realize you could chain spread operations...

```js
    await octokit.rest.pulls.update({
      pull_number: pullRequest.number,
      title: finalPrTitle,
      body: prBody,
      ...(prDraft ? {} : { "state": "open" }),
      ...github.context.repo,
    });
```

momentmaker: oh shoot - didn't refresh to see this. will create another pr for this.

PR #303, opened 2024-04-11
llm-pr-writer: add git messages for extra context

aelmanaa:


momentmaker:
could you add a changeset when you are ready with `pnpm changeset`?

PR #324, opened 2024-04-22
 feat: gha workflow validator 

erikburt:
### Summary

Adding a Github Actions workflow validation action.

This should be invoked on PRs, and it will fetch the diff of the PR and the target branch,  parse all the changes. It will then check if any changes are to workflow files, if a `uses: <action>` line appears it will validate 3 things.
1. That it uses a SHA1 reference
2. It has a proceeding comment, hopefully for the version/tag of the SHA1
3. If it uses node, it uses node20.

### Testing

Unit tests have decent coverage for a lot of the functionality.

Also did a lot of tests here: https://github.com/smartcontractkit/releng-test/pull/30

### Notes 

RE-2532


actions/gha-workflow-validator/jest.config.ts
```diff
@@ -0,0 +1,13 @@
+import {Config} from 'jest'
+export default {
+  displayName: "gha-workflow-validator",
+  preset: "../../jest.preset.js",
+  testEnvironment: "node",
+  transform: {
+    "^.+\\.[tj]s$": "@swc/jest",
+  },
+  moduleFileExtensions: ["ts", "js", "html"],
```
HenryNguyen5, line 9:
```suggestion
  moduleFileExtensions: ["ts"],
```
Can we remove these?

actions/gha-workflow-validator/jest.config.ts
```diff
@@ -0,0 +1,13 @@
...
+  coverageDirectory: "../../coverage/actions/gha-workflow-validator",
+  testPathIgnorePatterns: ["src/__tests__/__fixtures__",  "src/__tests__/__snapshots__/", "src/__tests__/__helpers__"],
+  collectCoverage: true,
```
HenryNguyen5, line 12: I'd probably put this behind a flag or use the `--coverage` flag instead, its faster to re-run tests without coverage

actions/gha-workflow-validator/jest.config.ts
```diff
@@ -0,0 +1,13 @@
+import {Config} from 'jest'
+export default {
+  displayName: "gha-workflow-validator",
+  preset: "../../jest.preset.js",
+  testEnvironment: "node",
+  transform: {
+    "^.+\\.[tj]s$": "@swc/jest",
```
HenryNguyen5, line 7:
```suggestion
    "^.+\\.ts$": "@swc/jest",
```
Do we need to do any js transforms?

actions/gha-workflow-validator/jest.config.ts
erikburt: The tests execute in about 0.4 seconds, I think it's ok to leave it enabled and it can be disabled if runtime of tests/coverage inflate.

actions/gha-workflow-validator/jest.config.ts
erikburt: Seems like it's best to include if we have to include js for `moduleFileExtensions`?

actions/gha-workflow-validator/jest.config.ts
erikburt: I can remove `html`

```
  >  NX   ● Validation Error:

   
     moduleFileExtensions must include 'js':
     but instead received:
       ["ts"]
     Please change your configuration to include 'js'.
   
     Configuration Documentation:
     https://jestjs.io/docs/configuration
```

actions/gha-workflow-validator/src/action-reference-validations.ts
```diff
@@ -0,0 +1,95 @@
...
+  return `${change.ref} is not a valid SHA1`;
+}
+
+function hasVersionComment(change: ActionReference) {
+  if (change.comment) return;
+
+  return `No version comment found`
+}
+
+async function isNode20Action(ghClient: Octokit, change: ActionReference) {
```
momentmaker, line 74: should be future-proof this by setting the `20` as a static var like `CURRENT_NODE_VERSION` or something?

momentmaker (approved):
nice! looks great!

actions/gha-workflow-validator/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+import { commentOnPrOrUpdateExisting, deleteCommentOnPRIfExists, getComparison } from "./github.js";
+import { validateActionReferenceChanges } from "./action-reference-validations.js";
+import { filterForGithubWorkflowChanges, formatGithubComment, parseAllAdditions } from "./utils.js";
+import { COMMENT_HEADER } from "./strings.js";
+
+(async () => {
+  const { token, owner, repo, base, head, prNumber } = getInvokeContext();
+  const octokit = github.getOctokit(token);
+
+  const allFiles = await getComparison(octokit, owner, repo, base, head);
```
HenryNguyen5, line 12: How do you feel about this?

```suggestion
  const changedFiles = await getChangedFiles(octokit, owner, repo, base, head);
```

I'm wondering if we should accept a list of changed files instead, and chain it with something like https://github.com/tj-actions/changed-files

actions/gha-workflow-validator/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  const ghaWorkflowFiles = filterForGithubWorkflowChanges(allFiles);
```
HenryNguyen5, line 13: Same as above, we could use the aforementioned GHA to delegate this logic

actions/gha-workflow-validator/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  const containsWorkflowModifications = ghaWorkflowPatchAdditions.some(file => {
+    return (file.filename.startsWith('.github/workflows') || file.filename.startsWith('.github/actions')) && (file.filename.endsWith('.yml') || file.filename.endsWith('.yaml'))
+  });
+
+  if (containsWorkflowModifications) {
+    const actionReferenceValidations = await validateActionReferenceChanges(octokit, ghaWorkflowPatchAdditions)
+    const validationFailed = actionReferenceValidations.some(validation => validation.lineValidations.length > 0);
+    const invokedThroughPr = prNumber !== undefined;
+
+    if (validationFailed && invokedThroughPr) {
```
HenryNguyen5, line 25:
```suggestion
    if (!validationFailed) return;
```

Would be good if we can reduce nesting where we can

actions/gha-workflow-validator/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const allFiles = await getComparison(octokit, owner, repo, base, head);
+  const ghaWorkflowFiles = filterForGithubWorkflowChanges(allFiles);
+  const ghaWorkflowPatchAdditions = parseAllAdditions(ghaWorkflowFiles);
+
+  const containsWorkflowModifications = ghaWorkflowPatchAdditions.some(file => {
+    return (file.filename.startsWith('.github/workflows') || file.filename.startsWith('.github/actions')) && (file.filename.endsWith('.yml') || file.filename.endsWith('.yaml'))
+  });
+
+  if (containsWorkflowModifications) {
```
HenryNguyen5, line 20:
```suggestion
  if (!containsWorkflowModifications) return
```
Would be good if we can reduce nesting where we can

actions/gha-workflow-validator/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+    const actionReferenceValidations = await validateActionReferenceChanges(octokit, ghaWorkflowPatchAdditions)
+    const validationFailed = actionReferenceValidations.some(validation => validation.lineValidations.length > 0);
+    const invokedThroughPr = prNumber !== undefined;
+
+    if (validationFailed && invokedThroughPr) {
+      const errorMessage = formatGithubComment(actionReferenceValidations, owner, repo, head);
+      commentOnPrOrUpdateExisting(octokit, owner, repo, prNumber, errorMessage);
```
HenryNguyen5, line 27: Missing await?

actions/gha-workflow-validator/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+      return core.setFailed("Errors found in workflow files. See comment on for details.");
+    } else if (validationFailed) {
+      // If the action is not invoked through a PR, we can't comment on the PR
+      return core.setFailed("Errors found in workflow files.");
+    } else if (!validationFailed && invokedThroughPr) {
+      deleteCommentOnPRIfExists(octokit, owner, repo, prNumber);
```
HenryNguyen5, line 33: I think we should use annotations over comments, they should(?) work on PRs and pushes, and integrate better w/ the review process. 

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#using-workflow-commands-to-access-toolkit-functions

actions/gha-workflow-validator/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+})().catch((err) => {
+  core.error("Uncaught Error - " + err);
+  core.setFailed(err.message);
```
HenryNguyen5, line 38: Not sure that we need this, we can let the process crash. GHA should still set a failure and spit out a stacktrace

actions/gha-workflow-validator/src/strings.ts
```diff
@@ -0,0 +1,57 @@
+export const COMMENT_HEADER = `### GHA Workflow Validator Results`
+
+const FIXING_ERRORS = `
```
HenryNguyen5, line 3: I would try to use github's job summaries: https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/
or the notice annotation here instead

actions/gha-workflow-validator/src/__tests__/github.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    const result = await getComparison(octokit, owner, repo, base, head);
+
+    expect(result?.length).toBeDefined();
+    expect(result?.length).toEqual(27);
+
+    nockDone();
+  });
+});
+
+// todo
```
HenryNguyen5, line 81:
```suggestion
```

actions/gha-workflow-validator/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "gha-workflow-validator",
+  "version": "0.0.0",
```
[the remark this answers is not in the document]
erikburt: Will get minor bumped by changesets.

actions/gha-workflow-validator/project.json
```diff
@@ -0,0 +1,46 @@
...
+    },
+    "build:internal": {
+      "executor": "@nx/esbuild:esbuild",
+      "outputs": ["{options.outputPath}"],
+      "defaultConfiguration": "production",
+      "options": {
+        "format": ["cjs"],
+        "thirdParty": true,
+        "outputPath": "actions/gha-workflow-validator/dist/",
+        "main": "actions/gha-workflow-validator/src/index.ts",
```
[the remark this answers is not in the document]
erikburt: I've tried messing around with the `nx` variables and haven't had much success. I will give it another shot because it is rather annoying having to run the test/build commands only in the root dir of the repo.

actions/gha-workflow-validator/src/__tests__/__helpers__/test-utils.ts
```diff
@@ -0,0 +1,16 @@
+import { getOctokit } from "@actions/github";
+import fetch from "node-fetch";
+
+export function getTestOctokit(nockbackMode: string) {
```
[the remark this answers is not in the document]
erikburt: It has a string union on the `setMode` method - I can try and reference that.

actions/gha-workflow-validator/src/__tests__/action-reference-validation.test.ts
```diff
@@ -0,0 +1,79 @@
...
+const nockBack = nock.back;
+// Set the fixture path and nockBack mode
+nockBack.fixtures = path.join(__dirname, "__fixtures__");
+
+// Change to 'lockdown' to use existing fixtures
+// Valid values = lockdown, record, wild, dryrun, update
+nockBack.setMode("record");
+
+if (nockBack.currentMode === "lockdown") {
+  nock.disableNetConnect();
```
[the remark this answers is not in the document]
erikburt: I did it because I thought I saw some behaviour where it was trying to connect even when in lockdown. I may have misread the console output though?

actions/gha-workflow-validator/src/__tests__/action-reference-validation.test.ts
```diff
@@ -0,0 +1,79 @@
...
+import { ParsedFile } from '../utils.js';
+
+// nock-back provides the recording and playback functionality
+const nockBack = nock.back;
+// Set the fixture path and nockBack mode
+nockBack.fixtures = path.join(__dirname, "__fixtures__");
+
+// Change to 'lockdown' to use existing fixtures
+// Valid values = lockdown, record, wild, dryrun, update
+nockBack.setMode("record");
```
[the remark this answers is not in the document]
erikburt: Ah yes it should be `lockdown` good catch. Will try and set this up.

actions/gha-workflow-validator/src/action-reference-validations.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+  if (node20ActionValidation) {
+    validationErrors.push({ message: node20ActionValidation });
+  }
+
+  return validationErrors;
+}
+
+function usesShaRef(change: ActionReference) {
+  const sha1Regex = /^[0-9a-f]{40}$/;
```
chainchad, line 62: I think there is experimental sha256 support for git. Wondering if we should make rename referenences to sha1 to just sha?

actions/gha-workflow-validator/src/action-reference-validations.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+type ValidationError = { message: string; };
+
+export async function validateActionReferenceChanges(octokit: Octokit, changes: ParsedFile[]): Promise<ValidationResult[]> {
+  core.debug(`Validating action reference changes, on ${changes.length} changes`)
+
+  const resultsPromise = changes.map(async (change) => {
+    const lineValidationPromises  = change.addedLines.filter(addedLine => addedLine.actionReference).map(async (line) => {
+      const validationErrors = (line.actionReference) ? await validateLine(octokit, line.actionReference) : [];
```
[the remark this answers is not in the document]
erikburt: There's eslint at the repository level but it's only setup for two things, and nothing uses prettier. I was thinking it should be done repo-wide.

actions/gha-workflow-validator/src/action-reference-validations.ts
```diff
@@ -0,0 +1,95 @@
...
+  const shaRefValidation = usesShaRef(line);
+  const versionCommentValidation = hasVersionComment(line);
+  const node20ActionValidation = await isNode20Action(octokit, line);
+
+  if (shaRefValidation) {
+    validationErrors.push({ message: shaRefValidation });
+  }
+  if (versionCommentValidation) {
+    validationErrors.push({ message: versionCommentValidation });
+  }
```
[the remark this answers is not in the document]
erikburt: `hasVersionComment` returns `string | undefined`. The string is populated when there's an error.

actions/gha-workflow-validator/src/index.ts
erikburt: `tj-actions/changed-files` will not return the same thing as the Github API `compare` endpoint.

I'm trying to only validate the changes introduced in a PR, and by using the `compare` endpoint you get the diff. I don't think it's reasonable for people have to overhaul a whole file due to linting errors unrelated to their change.

actions/gha-workflow-validator/src/index.ts
erikburt: There's logic below for negated case, and this wouldn't reduce nesting.

actions/gha-workflow-validator/src/index.ts
erikburt: Oh I didn't even know about these. This is way better.

actions/gha-workflow-validator/src/action-reference-validations.ts
erikburt: Yes that sounds good. Although this regex would break for a sha256 hash, I'll have to do a second check.

actions/gha-workflow-validator/src/index.ts
erikburt: I don't see the benefit of using an external action to handle the computation of changed files, and using that as input.

The goal of this action is to validate changes to workflow files and only getting a list of modified files kind of defeats that right? Because you still don't know what was changed in that file directly.

For that use-case, the GH API compare endpoint is better for this use-case isn't it? It returns all files modified and the patch for each file.

Unless you're suggesting the linting happens for the whole file?

actions/gha-workflow-validator/src/github.ts
```diff
@@ -0,0 +1,174 @@
...
+) {
+  core.debug(`Getting comments on PR ${prNumber}`);
+  return await octokit.paginate(octokit.rest.issues.listComments, {
+    owner: owner,
+    repo: repo,
+    issue_number: prNumber,
+  });
+}
+
+async function createComment(
```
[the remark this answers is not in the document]
erikburt: Namespaced

actions/gha-workflow-validator/src/github.ts
```diff
@@ -0,0 +1,174 @@
...
+  const yamlPath = join(repoPath, "action.yaml");
+
+  let actionFile = await getFileFromGithub(octokit, owner, repo, ymlPath, ref);
+  if (!actionFile) {
+    actionFile =  await getFileFromGithub(octokit, owner, repo, yamlPath, ref);
+  }
+  return actionFile;
+}
+
+export async function commentOnPrOrUpdateExisting(
```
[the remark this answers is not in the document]
erikburt: Changed

actions/gha-workflow-validator/src/__tests__/action-reference-validation.test.ts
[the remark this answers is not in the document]
erikburt: It doesn't seem like there's a great way to set environment variables when invoking through an nx target. Going to put the inlined unix env vars in the readme.

actions/gha-workflow-validator/project.json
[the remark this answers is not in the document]
HenryNguyen5: yeah, that's odd. you should be able to run nx commands within the pkg dir too. 
I tested in the GATI repo, I was able to run `pnpm nx build` in a package dir and it worked fine

actions/gha-workflow-validator/src/__tests__/action-reference-validation.test.ts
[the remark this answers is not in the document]
HenryNguyen5: yeah, seems like you'd either have to use `run-commands` executor so you can inline it in there, but then you'd be ditching the jest executor entirely.

Another option would be to make a custom executor that wraps the jest one to offer an arg to switch on, but that seems super overkill

actions/gha-workflow-validator/src/index.ts
HenryNguyen5: Good point, I didnt realize you're doing a line diff rather than a file diff

actions/gha-workflow-validator/src/index.ts
HenryNguyen5: Good point, I didnt realize you're doing a line diff rather than a file diff

actions/gha-workflow-validator/src/index.ts
erikburt: Added :+1:

actions/gha-workflow-validator/src/strings.ts
erikburt: Added summaries :+1:

actions/gha-workflow-validator/src/index.ts
erikburt: Remvoed

actions/gha-workflow-validator/src/index.ts
erikburt: Unnested

momentmaker (approved):

actions/gha-workflow-validator/project.json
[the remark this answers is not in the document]
erikburt: I'm going to create a follow-up task for this, as well as the prettier setup.

actions/gha-workflow-validator/src/utils.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  // trim past the "uses:" substring to get "<owner>/<repo><optional path>@<ref> # <optional comment>""
+  const trimmedLine = line.substring(line.indexOf(trimSubString) + trimSubString.length).trim();
+
+  if (trimmedLine.startsWith("./")) {
+    // Local action reference - do not validate these yet (TODO)
+    return;
+  }
+
+  const [ actionIdentifier, ...comment ] = trimmedLine.split("#");
```
HenryNguyen5, line 100:
```suggestion
  const [ actionIdentifier, ...commentParts ] = trimmedLine.split("#");
```

What does the comment look like? Is this to handle the case where the comment contains the # character too?

actions/gha-workflow-validator/src/utils.ts
```diff
@@ -0,0 +1,165 @@
...
+
+export async function setSummary(validationResults: ValidationResult[], fileUrlPrefix: string) {
+  const headerRows: TableRow[] = [
+    // Header Row
+    [
+      { data: "Filename", header: true },
+      { data: "Line Number", header: true },
+      { data: "Violations", header: true },
+    ],
+  ];
```
HenryNguyen5, line 131:
```suggestion
  const headerRow: TableRow = 
     [
      { data: "Filename", header: true },
      { data: "Line Number", header: true },
      { data: "Violations", header: true },
  ];
```

actions/gha-workflow-validator/src/utils.ts
```diff
@@ -0,0 +1,165 @@
...
+    return [
+      ...acc,
+      firstRowForFile,
+      ...errorCellTuples
+    ];
+
+  }, [] as TableRow[]);
+
+  await core.summary
+    .addTable([...headerRows, ...errorRows])
```
HenryNguyen5, line 161:
```suggestion
    .addTable([headerRow, ...errorRows])
```

actions/gha-workflow-validator/src/utils.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  const errorRows = validationResults.reduce<TableRow[]>((acc, curr) => {
+    const filename = curr.filename;
+
+    const errorCellTuples: TableCell[][]  = curr.lineValidations.map(line => {
+      const lineNumberCell: TableCell = { data: htmlLink(`${line.line.lineNumber}`, `${fileUrlPrefix}/${filename}#L${line.line.lineNumber}`) }
+      const violationsCell: TableCell = { data: line.validationErrors.map(error => error.message).join(", ") }
+
+      return [ lineNumberCell, violationsCell ];
+    })
```
HenryNguyen5, line 141:
```suggestion
    const errorCellTuples: TableCell[][]  = curr.lineValidations.map(({line}) => {
      const lineNumberCell: TableCell = { data: htmlLink(`${line.lineNumber}`, `${fileUrlPrefix}/${filename}#L${line.lineNumber}`) }
      const violationsCell: TableCell = { data: line.validationErrors.map(error => error.message).join(", ") }

      return [ lineNumberCell, violationsCell ];
    })
```

Apparently in go land this is called stuttering lmao

HenryNguyen5 (changes requested):

actions/gha-workflow-validator/src/utils.ts
erikburt: Yeah I'm spreading here in case the comment contains a `#`. This is mostly to get the part in front of the comment cleanly.

actions/gha-workflow-validator/src/utils.ts
erikburt: I'll remove the comment, idk what else is happening here?

actions/gha-workflow-validator/src/utils.ts
erikburt: This would break the code technically. I kinda want to figure out better typing/naming for these properties but I'm going to leave as is for now.

Can't destructure just to line because other properties are used.

```
line.line
line.validationErrors
```

actions/gha-workflow-validator/src/utils.ts
erikburt: Ah I see whats happening.

actions/gha-workflow-validator/src/utils.ts
```diff
@@ -0,0 +1,165 @@
...
+  if (usesIndex === -1) {
+    // Not an action reference
+    return;
+  }
+
+  // trim past the "uses:" substring to get "<owner>/<repo><optional path>@<ref> # <optional comment>""
+  const trimmedLine = line.substring(line.indexOf(trimSubString) + trimSubString.length).trim();
+
+  if (trimmedLine.startsWith("./")) {
+    // Local action reference - do not validate these yet (TODO)
```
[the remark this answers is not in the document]
erikburt: I think it's ok - if the action gets modified it will get linted. Just hard to validate references to local actions.

actions/gha-workflow-validator/src/utils.ts
```diff
@@ -0,0 +1,165 @@
...
+  ref: string;
+  comment?: string;
+  line: string;
+}
+
+export function filterForGithubWorkflowChanges(files: GithubFiles): GithubFiles {
+  return files?.filter(entry => {
+    return (entry.filename.startsWith('.github/workflows') || entry.filename.startsWith('.github/actions'))
+    && (entry.filename.endsWith('.yml') || entry.filename.endsWith('.yaml'))
+  })
```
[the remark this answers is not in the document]
erikburt: name is reserved keyword, but will destructure to filename.

actions/gha-workflow-validator/src/utils.ts
HenryNguyen5: ah didnt notice

HenryNguyen5 (approved):

PR #333, opened 2024-04-24
fix: tag-branch-check script

momentmaker:
This should fix this error about `fatal: ambiguous argument 'main^{commit}': unknown revision or path not in the working tree.`
https://github.com/smartcontractkit/mercury-pipeline/actions/runs/8635900445/job/23680380899#step:3:15

Tested here:
https://github.com/smartcontractkit/mercury-pipeline/actions/runs/8822235380/job/24219938510#step:3:15

### DIFF
--- a/actions/guard-tag-from-branch/scripts/tag-branch-check.sh
+++ b/actions/guard-tag-from-branch/scripts/tag-branch-check.sh
@@ -15,7 +15,7 @@ if [[ -z "${BRANCH_REGEX:-}" ]]; then
 fi
 
 # Get the sha of the tag's ancestor commit
-TAG_SHA=$(git rev-parse "${TAG}^{commit}")
+TAG_SHA=$(git rev-parse "origin/${TAG}^{commit}")
 echo "::debug::TAG_SHA: ${TAG_SHA}"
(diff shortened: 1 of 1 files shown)


erikburt (commented):
Is this script the problem or should the calling action update what it's passing to include the `origin/` ?

Edit:

Or should we try what's there and fallback to `origin/` if the original does not work?

I ask because I'm worried about breaking other workflows with this change.

momentmaker:
> Is this script the problem or should the calling action update what it's passing to include the `origin/` ?
> 
> Edit:
> 
> Or should we try what's there and fallback to `origin/` if the original does not work?
> 
> I ask because I'm worried about breaking other workflows with this change.

ahh good point. now that I think about it I think the issue is in the mercury-pipeline action:
```
          tag: ${{ github.ref_name }}
```
it should be:
```
          tag: ${{ inputs.tag || github.ref_name }}
```

momentmaker:
like this:
https://github.com/smartcontractkit/mercury-pipeline/actions/runs/8823733445/job/24224819294#step:3:12

erikburt:
Can this be closed then @momentmaker ?

momentmaker:
yeah sounds good - thanks!

PR #336, opened 2024-04-25
Generic Nix build action

ro-tex:
This PR adds a generic GHA that can build a Nix project and cache its artefacts on S3.

TODO: This PR should probably be adapted to augment the `setup-nix` action, instead of being a separate one.

Example usage:
```
      - name: Build and cache
        uses: smartcontractkit/.github/actions/nix-build
        with:
          cache-url: s3://${{ secrets.NIX_CACHE_S3_BUCKET }}?region=${{ secrets.AWS_REGION }}
          cache-privkey: ${{ secrets.NIX_BINARY_SIGNING_KEY }}
          aws-region: ${{ secrets.AWS_REGION }}
          role-to-assume: ${{ secrets.AWS_OIDC_IAM_ROLE_PUBLISH_ARN }}
          role-duration-seconds: ${{ secrets. AWS_ROLE_DURATION_SECONDS }}
```


actions/nix-build/action.yml
```diff
@@ -0,0 +1,142 @@
...
+  using: composite
+  steps:
+    - name: AWS credentials
+      uses: aws-actions/configure-aws-credentials@v4
+      with:
+        aws-region: ${{ inputs.aws-region }}
+        role-to-assume: ${{ inputs.role-to-assume }}
+        role-duration-seconds: ${{ inputs.role-duration-seconds }}
+        aws-access-key-id: ${{ inputs.aws-access-key-id }}
+        aws-secret-access-key: ${{ inputs.aws-secret-access-key }}
```
chainchad, line 72:
```suggestion
        aws-access-key-id: ${{ inputs.aws-access-key-id }}
        aws-secret-access-key: ${{ inputs.aws-secret-access-key }}
        mask-aws-account-id: true
```

Don't think we need access keys (see earlier comment) and we should mask the AWS account ID so it doesn't appear in logs.

chainchad (changes requested):
Looks good. Left some general feedback. Might want another reviewer more familiar with nix to review those specific steps though.

actions/nix-build/action.yml
```diff
@@ -0,0 +1,142 @@
...
+    description: "Nix cache binary private signing key - not used if setup-only=true"
+    required: false
+    default: ""
+
+  # aws inputs --------------------------
+  # see https://github.com/aws-actions/configure-aws-credentials/blob/main/action.yml
+  aws-region:
+    description: "AWS region"
+    required: false
+    default: us-west-2
```
momentmaker, line 26: could we make this required and not expose the default value? we usually pass this a secret.

actions/setup-nix/action.yml
```diff
@@ -104,6 +104,7 @@ runs:
         sudo ln -s ~/.aws $ROOT_PATH
 
     - uses: cachix/install-nix-action@8887e596b4ee1134dae06b98d573bd674693f47c # v26
+      if: inputs.enable-cachix == 'true'
```
momentmaker, line 107: if we are adding this then we should create a changeset file for this action too

actions/nix-build/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "nix-build",
+  "version": "0.0.1",
```
momentmaker, line 3: I think this should be `0.0.0` and once it's release it will be `0.1.0`

actions/nix-build/README.md
```diff
@@ -0,0 +1,3 @@
+# nix-build
+
+> Builds a nix project and uploads its build artefacts to a cache
```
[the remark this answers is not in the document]
ro-tex: I wish someone could explain to me why Americans decided to change a bunch of words in the English language...
Will change.

actions/nix-build/package.json
ro-tex: Thanks for this. I wasn't sure whether 0.0.0 was a valid version.

actions/nix-build/action.yml
```diff
@@ -0,0 +1,142 @@
...
+    required: false
+  role-duration-seconds:
+    description: Role duration in seconds. Default is one hour.
+    required: false
+  aws-access-key-id:
+    description: AWS Access Key ID. Provide this key if you want to assume a role using access keys rather than a web identity token.
+    required: false
+  aws-secret-access-key:
+    description: AWS Secret Access Key. Required if aws-access-key-id is provided.
+    required: false
```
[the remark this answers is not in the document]
ro-tex: Ah, that's a good point. I needed these while developing the action but now we can narrow it down.

actions/nix-build/action.yml
ro-tex: GitHub already does that for us, I believe - anything that comes from the secrets appears as `***` in the output of the action.

actions/setup-nix/action.yml
ro-tex: Added.

PR #353, opened 2024-05-02
feat: github-actions-updater

erikburt:
### Summary

Introducing a tool to check repositories for node12/node16 actions dependencies, and a method for updating them.

### DIFF
--- a/apps/actions-dependencies-updater/README.md
+++ b/apps/actions-dependencies-updater/README.md
@@ -1,0 +1,34 @@
+# Actions Dependencies Updater
+
+This tool has two main purposes:
+1. Checking if your workflows rely (directly or transitively) on any node12 or node16 actions.
+2. Updating all actions within a repo's workflows to the newest tag/release for that action.
+
+## Setup
+
+Have [`pnpm`](https://pnpm.io/) and [`direnv`](https://direnv.net/) setup.
+
+1. Install Dependencies
+
+```bash
+pnpm install
+```
+
+2. Populate a `.envrc` file within this project's directory
+
+```bash
+# Requires "repo" access, and authorized to read the repo(s) you are testing.
+export GH_ACCESS_TOKEN="<Github PAT>"
+```
+
+3.  `direnv allow`
+
+## Usage
+
+```
+pnpm start --help
+```
+
+### Example
+
+` pnpm start --repo-dir=/Users/erik/Documents/repos/chainlink`
--- a/apps/actions-dependencies-updater/README.md
+++ b/apps/actions-dependencies-updater/README.md
@@ -9,22 +9,23 @@ This tool has two main purposes:
 
 ## Setup
 
-Have [`pnpm`](https://pnpm.io/) and [`direnv`](https://direnv.net/) setup.
+1. Install and setup tooling
 
-1. Install Dependencies
+   - [`pnpm`](https://pnpm.io/)
+   - (optional) [`direnv`](https://direnv.net/)
 
-```bash
-pnpm install
-```
-
-2. Populate a `.envrc` file within this project's directory
-
-```bash
-# Requires "repo" access, and authorized to read the repo(s) you are testing.
-export GH_ACCESS_TOKEN="<Github PAT>"
-```
-
-3.  `direnv allow`
+2. Install Dependencies
+   ```bash
+    pnpm install
+   ```
+3. Generate a Github Personal Access Token with repo access
+   - This ideally has read access to all actions used in your workflow files
+4. If using `direnv`:
+   - Populate a `.envrc` file within this project's directory
+   ```bash
+   export GH_ACCESS_TOKEN="<Github PAT>"
+   ```
+   - `direnv allow`
 
 ## Usage
 
@@ -44,18 +45,21 @@ pnpm start --repo-dir=<path to repository>
   default) perform `git commit` for you.
   - If you have setup commit signing, you will be required to sign commits as
     you usually do.
+- If you're _not_ using `direnv` then you will need to export or inline the
+  `GH_ACCESS_TOKEN` environment variable.
+  - For example: `GH_ACCESS_TOKEN="<pat>" pnpm start ...`
 
 ### Examples
 
 - Update all actions, adn check for node12/node16 action dependencies before and
   after the update
 
-  - ` pnpm start --repo-dir=/Users/user/repos/chainlink`
+  - `pnpm start --repo-dir=/Users/user/repos/chainlink`
 
 - Update all actions, adn check for node12/node16 action dependencies before and
   after the update, but don't create a local branch, and don't commit anything
 
-  - ` pnpm start --repo-dir=/Users/user/repos/chainlink --no-branch --no-commit`
+  - `pnpm start --repo-dir=/Users/user/repos/chainlink --no-branch --no-commit`
 
 - Only check for deprecated action dependencies, and perform no updates
-  - ` pnpm start --repo-dir=/Users/user/repos/chainlink --check-deprecated`
+  - `pnpm start --repo-dir=/Users/user/repos/chainlink --check-deprecated`
--- a/apps/actions-dependencies-updater/README.md
+++ b/apps/actions-dependencies-updater/README.md
@@ -38,6 +38,8 @@ pnpm start --repo-dir=<path to repository>
 
 ### Notes
 
+- This makes no remote changes, all changes are local only. This means you are
+  required to push the branch and open a PR for these changes.
 - When updating action references, it will update them in batches and (by
   default) perform `git commit` for you.
   - If you have setup commit signing, you will be required to sign commits as
--- a/apps/actions-dependencies-updater/README.md
+++ b/apps/actions-dependencies-updater/README.md
@@ -1,8 +1,11 @@
 # Actions Dependencies Updater
 
 This tool has two main purposes:
-1. Checking if your workflows rely (directly or transitively) on any node12 or node16 actions.
-2. Updating all actions within a repo's workflows to the newest tag/release for that action.
+
+1. Checking if your workflows rely (directly or transitively) on any node12 or
+   node16 actions.
+2. Updating all actions within a repo's workflows to the newest tag/release for
+   that action.
 
 ## Setup
 
@@ -26,9 +29,31 @@ export GH_ACCESS_TOKEN="<Github PAT>"
 ## Usage
 
 ```
+# Help Menu
 pnpm start --help
+
+# Basic Execution
+pnpm start --repo-dir=<path to repository>
 ```
 
-### Example
+### Notes
 
-` pnpm start --repo-dir=/Users/erik/Documents/repos/chainlink`
+- When updating action references, it will update them in batches and (by
+  default) perform `git commit` for you.
+  - If you have setup commit signing, you will be required to sign commits as
+    you usually do.
+
+### Examples
+
+- Update all actions, adn check for node12/node16 action dependencies before and
+  after the update
+
+  - ` pnpm start --repo-dir=/Users/user/repos/chainlink`
+
+- Update all actions, adn check for node12/node16 action dependencies before and
+  after the update, but don't create a local branch, and don't commit anything
+
+  - ` pnpm start --repo-dir=/Users/user/repos/chainlink --no-branch --no-commit`
+
+- Only check for deprecated action dependencies, and perform no updates
+  - ` pnpm start --repo-dir=/Users/user/repos/chainlink --check-deprecated`
--- a/apps/actions-dependencies-updater/README.md
+++ b/apps/actions-dependencies-updater/README.md
@@ -12,7 +12,6 @@ This tool has two main purposes:
 1. Install and setup tooling
 
    - [`pnpm`](https://pnpm.io/)
-   - (optional) [`direnv`](https://direnv.net/)
 
 2. Install Dependencies
    ```bash
@@ -20,12 +19,10 @@ This tool has two main purposes:
    ```
 3. Generate a Github Personal Access Token with repo access
    - This ideally has read access to all actions used in your workflow files
-4. If using `direnv`:
-   - Populate a `.envrc` file within this project's directory
+4. Populate a `.env` file within this project's directory
    ```bash
-   export GH_ACCESS_TOKEN="<Github PAT>"
+   GH_ACCESS_TOKEN="<Github PAT>"
    ```
-   - `direnv allow`
 
 ## Usage
 
--- a/apps/actions-dependencies-updater/README.md
+++ b/apps/actions-dependencies-updater/README.md
@@ -61,5 +61,5 @@ pnpm start --repo-dir=<path to repository>
 
   - `pnpm start --repo-dir=/Users/user/repos/chainlink --no-branch --no-commit`
 
-- Only check for deprecated action dependencies, and perform no updates
-  - `pnpm start --repo-dir=/Users/user/repos/chainlink --check-deprecated`
+- Only check for deprecated action dependencies, perform no updates
+  - `pnpm start --repo-dir=/Users/user/repos/chainlink --skip-updates`
--- a/apps/actions-dependencies-updater/README.md
+++ b/apps/actions-dependencies-updater/README.md
@@ -51,12 +51,12 @@ pnpm start --repo-dir=<path to repository>
 
 ### Examples
 
-- Update all actions, adn check for node12/node16 action dependencies before and
+- Update all actions, and check for node12/node16 action dependencies before and
   after the update
 
   - `pnpm start --repo-dir=/Users/user/repos/chainlink`
 
-- Update all actions, adn check for node12/node16 action dependencies before and
+- Update all actions, and check for node12/node16 action dependencies before and
   after the update, but don't create a local branch, and don't commit anything
 
   - `pnpm start --repo-dir=/Users/user/repos/chainlink --no-branch --no-commit`
--- a/apps/actions-dependencies-updater/README.md
+++ b/apps/actions-dependencies-updater/README.md
@@ -12,6 +12,8 @@ This tool has two main purposes:
 1. Install and setup tooling
 
    - [`pnpm`](https://pnpm.io/)
+   - [`asdf`](https://asdf-vm.com/), then
+     [`nodejs`](https://github.com/asdf-vm/asdf-nodejs).
 
 2. Install Dependencies
    ```bash
--- a/apps/actions-dependencies-updater/package.json
+++ b/apps/actions-dependencies-updater/package.json
@@ -14,6 +14,7 @@
   "license": "MIT",
   "repository": "https://github.com/smartcontractkit/.github",
   "dependencies": {
+    "dotenv": "^16.4.5",
     "minimist": "^1.2.8",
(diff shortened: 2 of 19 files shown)


apps/actions-dependencies-updater/README.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+2. Populate a `.envrc` file within this project's directory
+
+```bash
+# Requires "repo" access, and authorized to read the repo(s) you are testing.
+export GH_ACCESS_TOKEN="<Github PAT>"
+```
+
+3.  `direnv allow`
```
HenryNguyen5: If we use https://www.npmjs.com/package/dotenv then we dont have to ask the user to have another system level package installed

apps/actions-dependencies-updater/README.md
```diff
@@ -0,0 +1,61 @@
...
+- This makes no remote changes, all changes are local only. This means you are
+  required to push the branch and open a PR for these changes.
+- When updating action references, it will update them in batches and (by
+  default) perform `git commit` for you.
+  - If you have setup commit signing, you will be required to sign commits as
+    you usually do.
+
+### Examples
+
+- Update all actions, adn check for node12/node16 action dependencies before and
```
HenryNguyen5, line 54:
```suggestion
- Update all actions, and check for node12/node16 action dependencies before and
```

apps/actions-dependencies-updater/README.md
```diff
@@ -0,0 +1,61 @@
...
+  after the update
+
+  - ` pnpm start --repo-dir=/Users/user/repos/chainlink`
+
+- Update all actions, adn check for node12/node16 action dependencies before and
```
HenryNguyen5, line 59:
```suggestion
- Update all actions, and check for node12/node16 action dependencies before and
```

apps/actions-dependencies-updater/src/caches.mts
```diff
@@ -0,0 +1,95 @@
+import * as log from "./logger.mjs";
+import { GithubShaToVersionCache } from "./github.mjs";
+import { UpdateTransaction } from "./updater.mjs";
+
+const CACHE_DIR = ".cache/";
```
HenryNguyen5, line 5: So this will create a `.cache` in the package directory, since we're depending on `pnpm` for execution?

HenryNguyen5, line 5:
```suggestion
const CACHE_DIR = ".cache";
```

`join` will add path seperators for you

apps/actions-dependencies-updater/src/caches.mts
```diff
@@ -0,0 +1,95 @@
...
+const SHA_TO_VERSION_CACHE = "sha-to-version.json";
+
+export function initialize(forceRefresh: boolean, optionalId?: string) {
+  const id = optionalId ? optionalId : Date.now();
```
HenryNguyen5, line 9:
```suggestion
export function initialize(forceRefresh: boolean, id = Date.now()) {

```

apps/actions-dependencies-updater/src/caches.mts
```diff
@@ -0,0 +1,95 @@
...
+  const cacheExists = fs.existsSync(path.join(CACHE_DIR, SHA_TO_VERSION_CACHE));
+
+  if (cacheExists && forceRefresh) {
+    try {
+      // rename .cache to .cache-<timestamp>
```
HenryNguyen5, line 14: Why don't we delete the cache rather than keeping the old one? It shouldn't be important to recover it since its cache data.

apps/actions-dependencies-updater/src/caches.mts
```diff
@@ -0,0 +1,95 @@
...
+      const newCachePath = path.join(
+        CACHE_DIR,
+        SHA_TO_VERSION_CACHE.replace(".json", `-${id}.json`),
+      );
+      log.warn(
+        `Forcing cache refresh. Previous cache will be moved to ${newCachePath}`,
+      );
+      fs.renameSync(path.join(CACHE_DIR, SHA_TO_VERSION_CACHE), newCachePath);
+    } catch (e) {
+      log.error(`Failed to remove cache: ${e}`);
```
HenryNguyen5, line 24:
```suggestion
      log.error(`Failed to refresh cache: ${e}`);
```

apps/actions-dependencies-updater/src/caches.mts
```diff
@@ -0,0 +1,95 @@
...
+    ),
+    updateTransactions: new Cache<UpdateTransaction>(
+      false,
+      `updates-${id}.json`,
+      {},
+    ),
+  };
+}
+
+class Cache<T> {
```
HenryNguyen5, line 41:
```suggestion
class Cache<T extends Record<string, any>> {
```

apps/actions-dependencies-updater/src/caches.mts
```diff
@@ -0,0 +1,95 @@
...
+
+  constructor(initializeFromDisk: boolean, filePath: string, initial?: T) {
+    this.filePath = path.join(CACHE_DIR) + filePath;
+
+    if (initializeFromDisk) {
+      this.cache = this.load();
+    } else if (initial) {
+      this.cache = initial;
+    } else {
+      this.cache = {} as T;
```
HenryNguyen5, line 53:
```suggestion
      this.cache = {}
```

We might not need this cast if we extend the generic

apps/actions-dependencies-updater/src/caches.mts
```diff
@@ -0,0 +1,95 @@
...
+
+  public get(): T {
+    return this.cache;
+  }
+
+  public set(key: keyof T, value: T[keyof T]) {
+    this.cache[key] = value;
+  }
+
+  public getValue(key: keyof T, ...keys: (keyof T)[]): T[keyof T] {
```
HenryNguyen5, line 65:
```suggestion
  public getValue(key: keyof T): T[keyof T] {
```

apps/actions-dependencies-updater/src/git-cli.mts
```diff
@@ -0,0 +1,69 @@
...
+
+export async function commit(ctx: RunContext, message: string) {
+  if (!ctx.git.commit) {
+    log.debug("Skipping commit (no-commit flag set)");
+    return;
+  }
+
+  await cdWrap(ctx.repoDir, async () => {
+    log.info(`Committing changes with message: ${message}`);
+    log.info("Tap your YubiKey!");
```
HenryNguyen5, line 56:
```suggestion
```
Shouldn't passportcontrol have a notif for this?

apps/actions-dependencies-updater/src/git-cli.mts
```diff
@@ -0,0 +1,69 @@
...
+  });
+}
+
+async function cdWrap(repositoryDirectory: string, fxn: () => Promise<void>) {
+  const prevDir = (await $`pwd`).stdout.trim();
+  cd(repositoryDirectory);
+
+  await fxn().finally(() => {
+    cd(prevDir);
+  });
```
HenryNguyen5, line 68: Prefer `pushd/popd` since it's designed for this use case. You should be able to drop the `.finally` with `popd` too.

apps/actions-dependencies-updater/src/git-cli.mts
```diff
@@ -0,0 +1,69 @@
...
+
+  await cdWrap(ctx.repoDir, async () => {
+    log.info(`Committing changes with message: ${message}`);
+    log.info("Tap your YubiKey!");
+    await $`git add --all`;
+    await $`git commit -m ${message}`;
+  });
+}
+
+async function cdWrap(repositoryDirectory: string, fxn: () => Promise<void>) {
```
HenryNguyen5, line 62:
```suggestion
async function performInDir(repositoryDirectory: string, fn: () => Promise<void>) {
```
1. WDYT about this name?
2. Typically callbacks are called `fn` or `cb`.

apps/actions-dependencies-updater/src/utils.mts
```diff
@@ -0,0 +1,166 @@
...
+      versions = versions.filter((v) => v.prefix === actionName);
+    }
+  }
+
+  // Sort the versions by comparing major, minor, and patch numbers
+  versions.sort((a, b) => {
+    if (a.major !== b.major) return a.major - b.major;
+    if (a.minor !== b.minor) return a.minor - b.minor;
+    return a.patch - b.patch;
+  });
```
HenryNguyen5, line 51: We should use semver sort instead https://www.npmjs.com/package/semver

apps/actions-dependencies-updater/src/git-cli.mts
```diff
@@ -0,0 +1,69 @@
+import * as log from "./logger.mjs";
```
HenryNguyen5, line 1: I found https://www.npmjs.com/package/simple-git to be pretty nice when I was working with signed changesets

apps/actions-dependencies-updater/src/generated/types/index.mts
```diff
@@ -0,0 +1,4 @@
+import { MySchema as WorkflowSchema } from "./github-workflow.mjs";
+import { MySchema as ActionSchema } from "./github-action.mjs";
+
+export { ActionSchema, WorkflowSchema };
```
HenryNguyen5, line 4:
```suggestion
export { MySchema as WorkflowSchema } from "./github-workflow.mjs";
export { MySchema as ActionSchema } from "./github-action.mjs";
```
This _might_ work, I'm a little rusty on export syntax

apps/actions-dependencies-updater/src/updater.mts
```diff
@@ -0,0 +1,192 @@
...
+) {
+  // iterate over workflows
+  for (const workflow of Object.values(workflowsByName)) {
+    log.debug(`Workflow: ${workflow.name} - path: ${workflow.path}`);
+
+    for (const job of workflow.jobs) {
+      log.debug(`Job: ${job.name}`);
+
+      for (const dependency of job.directDependencies) {
+        await processActionDependency(ctx, dependency, workflow.path);
```
HenryNguyen5, line 41: Do we have to worry about this being serially executed? If so, we could process this concurrently instead via `Promise.all`

apps/actions-dependencies-updater/src/updater.mts
```diff
@@ -0,0 +1,192 @@
...
+  filePath: string,
+  existingRef: string,
+  newRef: string,
+  newVersion: string,
+  innerRepoPath?: string,
+) {
+  const cacheKey = `${owner}/${repo}${innerRepoPath || ""}`;
+
+  if (!ctx.caches.updateTransactions.exists(cacheKey)) {
+    ctx.caches.updateTransactions.set(cacheKey, {
```
HenryNguyen5, line 134: Should we consider a `setIfExists` cache method?

HenryNguyen5 (changes requested):

apps/actions-dependencies-updater/src/caches.mts
erikburt: It's nice to cache this if you're running it for multiple repos in succession. The longest part of the execution is fetching from github.

I don't know what you mean by "since we're depending on pnpm for execution".

apps/actions-dependencies-updater/src/caches.mts
erikburt: Probably true for most people, I've found it useful.

apps/actions-dependencies-updater/src/caches.mts
erikburt: Still need it.

```
{} is assignable to the constraint of type  T  , but T could be instantiated with a different subtype of constraint  Record<string, any>  .
```

apps/actions-dependencies-updater/src/git-cli.mts
erikburt: I don't get notifications, and it may not be obvious to a user.

apps/actions-dependencies-updater/src/caches.mts
erikburt: That's what I thought too - but when I remove this slash I get this file in my package dir: `.cachesha-to-version.json`

apps/actions-dependencies-updater/src/git-cli.mts
erikburt: What I have works well. I found there's problems with libraries not respecting the env vars needed for commit signing.

apps/actions-dependencies-updater/src/generated/types/index.mts
erikburt: I'll leave it as simple and easy to read - an extra line makes it more obvious what is going on imo.

apps/actions-dependencies-updater/src/git-cli.mts
erikburt: On a bit of a time crunch and don't want to switch to something when I have something that also works.

apps/actions-dependencies-updater/src/utils.mts
erikburt: There's a lot of nice utilities in that package. `semver.coerce` would've saved me a good amount of time lmao

apps/actions-dependencies-updater/README.md
erikburt: Good point

apps/actions-dependencies-updater/src/caches.mts
```diff
@@ -0,0 +1,95 @@
...
+    ),
+  };
+}
+
+class Cache<T> {
+  private filePath: string;
+  private cache: T;
+
+  constructor(initializeFromDisk: boolean, filePath: string, initial?: T) {
+    this.filePath = path.join(CACHE_DIR) + filePath;
```
erikburt, line 46: oh, this is my problem

apps/actions-dependencies-updater/src/updater.mts
erikburt: Added concurrency.

apps/actions-dependencies-updater/src/updater.mts
erikburt: Added `getValueOrDefault` - ac12d5a30f1e83fcaa0172b19487b7e25316d4b2

apps/actions-dependencies-updater/src/git-cli.mts
erikburt: Looks good :+1:

apps/actions-dependencies-updater/src/utils.mts
erikburt: Using - 50cc1960ceed8ebe36a118d48cc38d39c5827fb2

apps/actions-dependencies-updater/src/caches.mts
HenryNguyen5: Using pnpm for executing the script will always set the cwd to where the package manifest lives. I was asking just to confirm that, since usually we'll package up something like this with https://github.com/vercel/ncc

apps/actions-dependencies-updater/src/workflows.mts
```diff
@@ -0,0 +1,401 @@
...
+  ctx: RunContext,
+  workflow: string,
+  jobKey: string,
+  jobDefinition: WorkflowSchema["jobs"]["key"],
+): Promise<Job> {
+  log.debug(`Processing Job: ${workflow} - ${jobKey}`);
+
+  if ("uses" in jobDefinition) {
+    log.warn(
+      "Found Job which is uses a workflow_call, to initiate another workflow. Parsing this is not directly supported, but if the workflow is in the same repository it should be parsed regularly.",
```
HenryNguyen5, line 151: I don't really understand this log, could you clarify it?

apps/actions-dependencies-updater/src/workflows.mts
```diff
@@ -0,0 +1,401 @@
...
+      name: jobKey,
+      containingWorkflow: workflow,
+      directDependencies: [],
+    };
+  }
+
+  if (!("steps" in jobDefinition)) {
+    log.warn(
+      "No steps found in job definition, bad job definition? Assuming no dependencies",
+    );
```
HenryNguyen5, line 163: This should error, no steps creates an invalid workflow

apps/actions-dependencies-updater/src/caches.mts
erikburt: We can also do that - what do you think is the best way?

apps/actions-dependencies-updater/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "lib": ["es2023"],
+    "module": "node16",
```
HenryNguyen5, line 5: We're using node20 in `.tool-versions`

apps/actions-dependencies-updater/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "actions-dependencies-updater",
+  "version": "0.0.0",
```
HenryNguyen5, line 3: It'd be good to put an `engine` check in here for `node20`, especially since we'll have others using this script locally.

HenryNguyen5 (changes requested):

apps/actions-dependencies-updater/src/caches.mts
HenryNguyen5: I wouldn't bother with binary packaging if there's an urge to get this running. It can cause its own problems when it comes to building for different platforms and binary distributions. If this becomes high demand we should probably merge this into `clunk` instead which already has a CLI framework and binary distribution setup

apps/actions-dependencies-updater/src/workflows.mts
erikburt: A workflow job can be another workflow definition (which has `workflow_call`) trigger. I don't really want to support that kind of relationship.

But if a job is essentially referencing another workflow, then that workflow it is referencing should be parsed and validated normally like all of the other workflows, and so we don't need to recurse from a workflow->job->workflow.

I don't know how to word that better? I'll throw it into a chatgpt and see I guess. This is also an edge-case I've seen only once.

apps/actions-dependencies-updater/src/workflows.mts
erikburt: This script doesn't care. If there's no steps then no problem, there's no updates.

apps/actions-dependencies-updater/tsconfig.json
erikburt: https://www.npmjs.com/package/@tsconfig/node20

apps/actions-dependencies-updater/src/workflows.mts
erikburt: `"Job detected that triggers another workflow via 'workflow_call'. Direct parsing of this workflow-to-workflow relationship isn't supported. However, if the called workflow resides in the same repository, it will be parsed as a standard workflow."`

apps/actions-dependencies-updater/tsconfig.json
erikburt: https://www.typescriptlang.org/docs/handbook/modules/theory.html#the-module-output-format

HenryNguyen5 (approved):

PR #358, opened 2024-05-03
feat: actions-updater workflow_call, and caching improvements

erikburt:
This introduces three new features to the `actions-dependencies-updater`

1. Partial support for `workflow_call` jobs, that is jobs which invoke a workflow rather than have steps
2. A `--skip-checks` flag to skip the recursive dependency checks
3. Debug information tracking for requests to github due to rate limits
4. Cache parsed actions to speed up processing (immutable references only)
5. A runtime cache to track which dependencies are direct, to limit unnecessary requests to github

### Testing

Will test on some repos once I'm no longer rate limited.

### DIFF
--- a/apps/actions-dependencies-updater/src/github.mts
+++ b/apps/actions-dependencies-updater/src/github.mts
@@ -163,6 +163,7 @@ async function getAllTags(
   repo: string,
 ): Promise<GitVersionTag[]> {
   try {
+    ctx.debug.tagRequests++;
     const response = (await ctx.octokit.request(
       "GET /repos/{owner}/{repo}/git/matching-refs/tags",
       { owner, repo },
@@ -207,6 +208,7 @@ async function getCommitForAnnotatedTag(
   apiPath: string,
 ): Promise<string | undefined> {
   try {
+    ctx.debug.tagRequests++;
     const response = (await ctx.octokit.request(
       "GET " + apiPath,
     )) as GetTagResponse;
@@ -243,7 +245,7 @@ export async function getActionFile(
   const yamlPath = join(repoPath, "action.yaml");
 
   let actionFile = await getFile(
-    ctx.octokit,
+    ctx,
     owner,
     repo,
     ymlPath,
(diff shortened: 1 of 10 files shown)


apps/actions-dependencies-updater/src/github.mts
```diff
@@ -261,18 +261,24 @@ export async function getActionFile(
...
   repo: string,
   path: string,
   ref: string,
 ) {
   try {
+    if (path.startsWith("/")) path = path.substring(1);
+    if (path.startsWith("./")) path = path.substring(2);
+
+    if (path === "") {
+      log.warn("Empty path provided to github.getFile. Returning empty.");
```
chainchad, line 276: Just confirming if this should be an error instead?

apps/actions-dependencies-updater/src/workflows.mts
```diff
@@ -398,3 +391,85 @@ function extractActionIdentifiersFromSteps(
     .filter((s): s is { uses: string } => !!s.uses)
     .map(({ uses }: { uses: string }) => uses);
 }
+
+/**
+ * An edge-case for parsing jobs which invoke a workflow_call. Which essentially references another workflow file, local or remote.
+ */
+async function parseWorkflowCall(
```
chainchad, line 398: Will this or should this catch using other workflows like this?

```
jobs:
  call_remote_workflow:
    uses: org-name/repo-name/.github/workflows/remote-workflow.yml@main
    with:
      example_input: 'Hello from another repo!'
    secrets:
      EXAMPLE_SECRET: ${{ secrets.EXAMPLE_SECRET }}
```

erikburt: Yeah, that's exactly what this method is designed for.

It's getting kinda messy to track actions/workflows local to an external action/workflow, so this method is kind of stubbed. This is something I should deal with if this tool sees more uses, but for now it's unfortunately a little janked given the timeline.

chainchad (approved):

apps/actions-dependencies-updater/src/github.mts
erikburt: Upgraded to error - a lot of the levels are all over the place to not overload people who are using the tool.

chainchad (approved):

momentmaker (approved):

PR #375, opened 2024-05-08
Avoid using graphql in gh cli due to intermittent error

chainchad:


actions/cleanup-old-branches/scripts/cleanup-old-branches.sh
```diff
@@ -156,66 +148,25 @@ function should_keep_branch() {
...
-    end_cursor=$(echo "$response" | jq -r '.data.repository.pullRequests.pageInfo.endCursor')
-  done
+  local pr_branch_names
+  # 1,000 PRs is the hard limit.
+  pr_branch_names=$(gh pr list --state open \
+    --json headRefName --jq '.[].headRefName' --limit 1000)
+  readarray -t OPEN_PR_BRANCHES <<< "${pr_branch_names}"
+
+  if [[ "${#OPEN_PR_BRANCHES[@]}" -eq 1000 ]]; then
+    echo "::error::Too many open PR Branches. Limit of 1,000 open PRs reached: ${OPEN_PR_BRANCHES[*]}"
```
HenryNguyen5, line 161: Should we really exit here? I feel like it should be fine to just batch process and warn that not all branches were processed.

HenryNguyen5 (approved):

actions/cleanup-old-branches/scripts/cleanup-old-branches.sh
chainchad: Discussing in a side channel.

PR #418, opened 2024-06-03
RE-1943: Add optional role session name input on setup-github-token

njegosrailic:
## What

Adding an optional role session name input to setup-github-token. It's an optional parameter, so it's not a breaking change.

## Why

I'm not sure if the proposed default value is the best one, but I'm still new and open to suggestions.

Ticket: https://smartcontract-it.atlassian.net/browse/RE-1943

## Test 

Tested from my branch: https://github.com/smartcontractkit/releng-test/actions/runs/9353197177

### DIFF
--- a/actions/setup-github-token/action.yml
+++ b/actions/setup-github-token/action.yml
@@ -15,10 +15,15 @@ inputs:
     description: Duration of role in seconds
     required: false
     default: "900"
+  role-session-name:
+    description: Session name to use when assuming the role.
+    required: false
+    default: "${{ github.run_id }}-${{ github.run_number }}-${{ github.job }}"
   set-git-config:
     description: Set git config
     required: false
     default: "false"
+
 outputs:
   access-token:
     value: ${{ steps.get-gh-token.outputs.access-token }}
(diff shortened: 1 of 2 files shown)


momentmaker (approved):

actions/setup-github-token/action.yml
```diff
@@ -27,13 +32,24 @@ outputs:
...
+
     - name: Assume role capable of getting token from gati
       uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
       with:
-        role-to-assume: ${{ inputs.aws-role-arn }}
-        role-duration-seconds: ${{ inputs.aws-role-duration-seconds }}
         aws-region: ${{ inputs.aws-region }}
         mask-aws-account-id: true
+        role-duration-seconds: ${{ inputs.aws-role-duration-seconds }}
+        role-session-name: ${{ inputs.role-session-name }}
```
erikburt, line 51: Won't this still use the potentially too long `${{ inputs.role-session-name }}`?

njegosrailic: We truncate it here to a maximum of 64 characters, regardless of whether the value is created using the default value or passed from another GH action:

https://github.com/smartcontractkit/.github/pull/418/files#diff-8b60da546acdf6951c641aa7ec7432c6923ff15b0d141f2db9ef309d3bb03be6R41

erikburt: That's not modifying the `${{ inputs.role-session-name }}` though is it?  You're piping that truncated value to `$GITHUB_OUTPUT`.

njegosrailic: That's right. I'll put a fix, thank you.

momentmaker: ahh yeah sorry - good catch!

PR #453, opened 2024-06-17
Optionally allow hotfix suffix in the monorepo tag

swladyka:
The change is backward compatible and optionally allows using semantic version compatible tag with the `-hotfix-<increment, e.g. timestamp>` suffix.

### DIFF
--- a/actions/check-git-tag-for-monorepo/package.json
+++ b/actions/check-git-tag-for-monorepo/package.json
@@ -1,6 +1,6 @@
 {
   "name": "check-git-tag-for-monorepo",
-  "version": "0.2.0",
+  "version": "0.1.0",
   "description": "checks if git tag is for a monorepo component <component>@<version>",
   "private": true,
   "scripts": {},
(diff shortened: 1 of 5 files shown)


actions/check-git-tag-for-monorepo/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "check-git-tag-for-monorepo",
-  "version": "0.1.0",
```
erikburt, line 3: Don't bump the version here directly, use `pnpm changeset add`.

erikburt (changes requested):

actions/check-git-tag-for-monorepo/package.json
swladyka: Hi Erik, thanks for the comment, I've corrected it. I suppose the changeset file once eventually merged will trigger the cicd-changeset action which takes care of versioning?

erikburt: Yes. Well merge the “Version packages” PR which will publish tags.

erikburt (approved):

PR #463, opened 2024-06-19
adding token as output

yashnevatia:


tt-cll (approved):
we're good to merge this yeah?

PR #469, opened 2024-06-25
fix: update ci-test-ts to run nx test targets

manjunathb4461:
Reverting the changes to `ci-test-ts` as the steps are the same as in `ci-lint-ts` to run `nx` test targets. 

I've tested the commands on https://github.com/smartcontractkit/releng-ts-app and they both work good.

### DIFF
--- a/actions/ci-test-ts/action.yml
+++ b/actions/ci-test-ts/action.yml
@@ -15,10 +15,6 @@ inputs:
     description: ""
     required: false
     default: "false"
-  base-ref:
-    description: ""
-    required: false
-    default: main
   # nodejs inputs
   node-version-file:
     description: ""
@@ -66,15 +62,14 @@ runs:
         run-install: "true"
       continue-on-error: true
 
-    - name: Run lint
-      if: inputs.check-only-affected != 'true'
+    - name: Run test
       shell: bash
-      run: pnpm run lint
+      run: pnpm run test
 
     - name: Run lint affected
       if: inputs.check-only-affected == 'true'
       shell: bash
-      run: pnpm run lint:affected --base=${{ inputs.base-ref }}
+      run: pnpm run test:affected
 
     - name: Collect metrics
       if: always() && inputs.metrics-job-name != ''
--- a/.changeset/young-roses-impress.md
+++ b/.changeset/young-roses-impress.md
@@ -1,0 +1,7 @@
+---
+"ci-test-ts": minor
+"nx-chainlink": patch
+---
+
+- update ci-test-ts to run correct pnpm commands
+- fix a unused variable error in `ns-chainlink`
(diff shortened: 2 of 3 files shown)


actions/ci-test-ts/action.yml
```diff
@@ -66,15 +66,14 @@ runs:
...
+    - name: Run test
       shell: bash
-      run: pnpm run lint
+      run: pnpm nx run-many --targets=test --verbose
 
     - name: Run lint affected
       if: inputs.check-only-affected == 'true'
       shell: bash
-      run: pnpm run lint:affected --base=${{ inputs.base-ref }}
+      run: pnpm nx affected --targets=test --verbose
```
chainchad, line 76: I think we can remove this input now if it's not being used

```
  base-ref:
    description: ""
    required: false
    default: main
```

chainchad (changes requested):

actions/ci-test-ts/action.yml
manjunathb4461: Removed it, thanks

actions/ci-test-ts/action.yml
```diff
@@ -66,15 +62,14 @@ runs:
         run-install: "true"
       continue-on-error: true
 
-    - name: Run lint
-      if: inputs.check-only-affected != 'true'
+    - name: Run test
       shell: bash
-      run: pnpm run lint
+      run: pnpm nx run-many --targets=test --verbose
```
momentmaker, line 67: hmm what if the project is not setup to use `nx`?

chainchad (approved):

momentmaker:
should we pin this to a specific/latest commit for this action?
https://github.com/smartcontractkit/.github/pull/469/files#diff-c00e54c6dcfdae524fff9eae7ae55eebcc79260b7b71ec61d948101e04d38c80L61

chainchad:
> 
should we pin this to a specific/latest commit for this action?
https://github.com/smartcontractkit/.github/pull/469/files#diff-c00e54c6dcfdae524fff9eae7ae55eebcc79260b7b71ec61d948101e04d38c80L61

Yeah, but there are probably other areas to update that too and that might be better served in a separate PR. @manjunathb4461 can you create a separate task to do that?

manjunathb4461:
@chainchad sure, I will create a ticket for it. Thanks!

chainchad (changes requested):
@manjunathb4461 don't forget to create changesets entries for the two packages that have changed.

manjunathb4461:
> @manjunathb4461 don't forget to create changesets entries for the two packages that have changed.

Sure, added the changeset for both the packages

chainchad (approved):

PR #472, opened 2024-06-26
RE-2527 add ci-sonarqube-ts action

manjunathb4461:
- add ci-sonarqube-ts action
- modify ci-lint-ts to upload eslint report
- modify ci-test-ts to upload lcov and junit reports


actions/ci-lint-ts/action.yml
```diff
@@ -76,6 +76,13 @@ runs:
       shell: bash
       run: pnpm run lint:affected --base=${{ inputs.base-ref }}
 
+    - name: Store lint result artifacts
+      if: always()
+      uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
+      with:
+        name: ts-lint-results
+        path: ./eslint-report.json
```
manjunathb4461, line 84: I don't see any of the repos using this action creating any format of report, so I'm assuming eslint as the standard and going with it

actions/ci-test-ts/action.yml
```diff
@@ -71,6 +71,15 @@ runs:
       shell: bash
       run: pnpm run test:affected
 
+    - name: Store test result artifacts
```
manjunathb4461, line 74: I don't see any of the repos using this action creating any format of report, so I'm assuming junit as the standard and going with it

actions/ci-lint-ts/action.yml
momentmaker: hmm maybe instead of running `pnpm lint` and storing the results (which I don't think it's doing?) we can use this action instead?

https://github.com/marketplace/actions/lint-action

it seems like it has annotation as well - what do you guys think @chainchad @erikburt

momentmaker: for eslint we will need to specify the output no?
https://eslint.org/docs/latest/use/command-line-interface

manjunathb4461: yes, we expect pnpm `lint` command in the repo using the action to provide the output, I've added an input parameter `eslint-json-report-paths` that they can use as well

erikburt: Should they be allowed to specify which lint command to run, and then also the expected path to the report?

manjunathb4461: I've come across some repos where the eslint report is generated but with a different artifact or file name, I've also added some defaults in case they don't want to use them (they wouldn't need to provide these inputs if they use our `ci-test-ts` and `ci-lint-ts`)

njegosrailic (approved):

actions/ci-lint-ts/action.yml
manjunathb4461: I'm resolving this conversation. if requested changes, I will implement them in separate PR

actions/ci-sonarqube-ts/action.yml
```diff
@@ -0,0 +1,126 @@
...
+        fetch-depth: ${{ inputs.checkout-repo-fetch-depth }}
+
+    - name: Download all workflow run artifacts
+      uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e # v4.1.7
+
+    - name: Set SonarQube Report Paths
+      id: sonarqube_report_paths
+      shell: bash
+      run: |
+        echo "sonarqube_junit_xml_report_paths=$(find inputs.test-report-artifact -name inputs.junit-xml-report-paths | paste -sd "," -)" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
manjunathb4461: Ahh yes, I hadn't noticed this issue in my tests earlier. Created a [PR](https://github.com/smartcontractkit/.github/pull/490) to fix this

PR #484, opened 2024-06-30
Group Dependencies in Renovate

manjunathb4461:
Started looking into this after discussions [here](https://chainlink-core.slack.com/archives/C03UV9B7M3L/p1719583508860899?thread_ts=1719571377.932819&cid=C03UV9B7M3L) 

- Renovate will group all minor and patch dependencies together by package manager name and create PRs with branch named after package manager `npm-dependencies`, `github-actions-dependencies` etc
- Major dependencies will continue to be raised separately per dependency

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -21,9 +21,14 @@
   },
   "packageRules": [
     {
-      "matchDepTypes": ["devDependencies"],
-      "matchUpdateTypes": ["patch"],
-      "enabled": false
+      "groupName": "major-minor dependencies grouped by manager",
+      "branchName": "{{manager}}-dependencies",
+      "matchUpdateTypes": ["major", "minor"]
+    },
+    {
+      "groupName": "all patch dependencies",
+      "branchName": "patch-dependencies",
+      "matchUpdateTypes": ["patch"]
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -200,6 +200,8 @@ importers:
 
   actions/nops-upgrade-checker: {}
 
+  actions/policy-bot-config-validator: {}
```
manjunathb4461, line 203: added by lefthook

.github/renovate.json
```diff
@@ -21,9 +21,14 @@
   },
   "packageRules": [
     {
-      "matchDepTypes": ["devDependencies"],
-      "matchUpdateTypes": ["patch"],
-      "enabled": false
+      "groupName": "major-minor dependencies grouped by manager",
+      "branchName": "{{manager}}-dependencies",
+      "matchUpdateTypes": ["major", "minor"]
```
momentmaker, line 26: it might make sense to not have `major` to be grouped as it probably has breaking changes

manjunathb4461: Should we group all minor and patch dependencies together per package manager and major updates can continue separately?

momentmaker: yeah that sounds good :)

manjunathb4461: updated accordingly, PTAL

momentmaker (approved):

PR #501, opened 2024-07-08
RE-2786 init go-mod-validator

manjunathb4461:
Implementation of code, action.yml and unit tests for go-mod-validator.

Tests
1. Validation failed - [chainlink-data-streams](https://github.com/smartcontractkit/chainlink-data-streams/actions/runs/9943503075)
2. Validation failed - [chainlink](https://github.com/smartcontractkit/crib/actions/runs/9942515397)
3. Validation successful - [crib](https://github.com/smartcontractkit/chainlink/actions/runs/9942521861)


apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,55 @@
...
+import { getSmartcontractkitDependencies } from "./parser";
+import { validateDependency } from "./github";
+
+async function run() {
+  // Access command-line arguments
+  const args = process.argv.slice(2);
+  if (args.length == 0) {
+    console.error("need 1 argument as the input");
+    process.exit(1);
+  }
```
erikburt, line 11: If this is to be used as a Github Action, I'd recommend looking at `@actions/core` for handling GHA inputs.

Some examples of nodejs actions:
- https://github.com/smartcontractkit/.github/blob/main/actions/signed-commits/
- https://github.com/smartcontractkit/.github/blob/main/actions/gha-workflow-validator/

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  // Read the file content
+  let goListOutputContent = "";
+  try {
+    goListOutputContent = fs.readFileSync(goListOutputFile, "utf-8");
+  } catch (err) {
+    console.error(
+      `failed to read input file: ${goListOutputFile}, with err: ${err}`,
+    );
+    process.exit(1);
```
erikburt, line 22: Example of `@actions/core` usage instead:

- `core.error("...")`
- `core.setFailed("...")`

apps/go-mod-validator/src/parser.ts
```diff
@@ -0,0 +1,76 @@
+const smartContractKitPrefix = "github.com/smartcontractkit/";
+
+function parseGoModLine(line: string) {
+  // Trim any extra whitespace
+  line = line.trim();
+
+  const moduleRegex = new RegExp(
+    /^([\w.-]+(?:\/[\w.-]+)*) (v\d+\.\d+\.\d+(?:-[\w\d.-]+)?(?:\+[\w\d.-]+)?)$/,
+  );
```
erikburt, line 9: Can you include a small comment so we can more easily infer what the regex is trying to accomplish.

apps/go-mod-validator/src/parser.ts
```diff
@@ -0,0 +1,76 @@
...
+      module,
+      version,
+    };
+  }
+
+  return null;
+}
+
+// extract all final dependencies
+function parseGoModLines(lines: string[]) {
```
erikburt, line 38: Would it make sense to use [`zx`](https://google.github.io/zx/getting-started) to invoke the `go list -m all` rather than having the output passed in as a file?

Could also do `go list -json -m all` to make it more usable during execution.

erikburt (changes requested):
Where will this be used? It seems like it's meant to operate during CI/CD so it should be an action. But it doesn't have the necessary `action.yml` file.

apps/go-mod-validator/src/parser.ts
manjunathb4461: I initially thought of running `go list -m all` in the script itself then we won't be able to use this script in envs where Go compiler isn't available. Second option I had on mind was to stream the output to the script like this `go list -m all | pnpm run start`, but I wasn't sure how difficult would testing the inputs for such a script be, so finally moved on to passing it as file input. Pls let me know if you think there are better options. 

I also checked `go list -json -m all`, it gives a weird json non-json format, where each module data is listed as json but no the entire output, parsing it was really difficult.

apps/go-mod-validator/src/index.ts
manjunathb4461: Sure, I will make those changes in this [ticket](https://smartcontract-it.atlassian.net/browse/RE-2786)

apps/go-mod-validator/src/parser.ts
manjunathb4461: Done, PTAL

apps/go-mod-validator/src/index.ts
manjunathb4461: One doubt I had about this is, we won't be able to use the script outside the github workflows. In case we want to use it pre-commit check or so in future. But if that's not a problem then I will definitely implement this with the `actions.yml` in the second PR

apps/go-mod-validator/src/parser.ts
erikburt: > won't be able to use this script in envs where Go compiler isn't available.

This script parses the output of `go list` right? When would the output be available, but the `go` command isn't?
If this is supposed to be run in CI/CD then the user should call `actions/setup-go` first, and if it isn't then you can list `go` as a requirement. 

> I also checked go list -json -m all, it gives a weird json non-json format, where each module data is listed as json but no the entire output, parsing it was really difficult.

Piping the output into `jq` wraps it into an array at least.

```
go list -json -m all | jq -s .
```

erikburt:
> Yes, there's another ticket for it. I will implement the github action in that ticket [smartcontract-it.atlassian.net/browse/RE-2786](https://smartcontract-it.atlassian.net/browse/RE-2786)

I think the initial implementation decisions should be made with this in mind, and I don't see the need to separate the two tasks. The user story should be concrete enough that you'll know if you're making a CLI tool, an action, or both. Based on the tasks above, it's either an action, or both.

If it's going to be used solely as an action, then build it out as a runnable action.

If its a CLI tool and action, then maybe it accepts command-line args (like you've done) but it has a composite `action.yml`  which can setup everything and run it like a GH action. In which case, I think you should assume that the running environment has `go` and `jq` installed, because the GH environment will.

erikburt:
Just a head's up, Github is terrible for PR reviewing when the branch was rebased. It breaks being able to see only the changes made since my last review.

I followed the rebase git flow before switching to CL as well but it makes it a nightmare for reviewing.

apps/go-mod-validator/src/parser.ts
manjunathb4461: yes that makes sense. I was thinking of keeping the `go` and `jq` details outside the script. Such that some steps are run before calling our script that actually take care of these details. I feel writing unit tests will also be easy for such scenarios.

manjunathb4461:
> The user story should be concrete enough that you'll know if you're making a CLI tool, an action, or both

As disucssed in the [ADR](https://docs.google.com/document/d/17PhnvIkV9sy19JsmCfKxYQh62R-m1tOa2bApxUod4OA/edit?disco=AAABQmkQNHY), we are going to implement only the github action. I should've mentioned it in the tickets clearly, my bad. 

But the reason I was thinking of keeping `go` and `jq` outside the script was to keep the script clean and make writing unit tests easier. But I will look into if it's worth the extra lines of code in `action.yml` and if not, I will move them into the script.

manjunathb4461:
> Just a head's up, Github is terrible for PR reviewing when the branch was rebased. It breaks being able to see only the changes made since my last review.

Was it because of the merge from main? I think it was not a rebase 🤔

erikburt:
> Was it because of the merge from main? I think it was not a rebase 🤔

Did you amend a commit then? Unfortunately any force push causes problems for reviewers. I see this:

> @manjunathb4461 [force-pushed](https://github.com/smartcontractkit/.github/compare/f3b5213b6f7b0ea8313f3dadeeaaaab419b53cbf..5afcbfc383e0ccfb1b4cb96aeebab1f1067b6ed0) the RE-2785-init-go-mod-validator branch from[ f3b5213 ](https://github.com/smartcontractkit/.github/commit/f3b5213b6f7b0ea8313f3dadeeaaaab419b53cbf)to 5afcbfc

manjunathb4461:
> force-pushed

ahh yes, I think it was bcoz of the amended commit, will try to avoid them. Thanks for pointing it out.

apps/go-mod-validator/src/parser.ts
manjunathb4461: I've moved `go list` and `jq` commands to the script. One more problems I noticed with redirecting the output of `go list | jq` to the script but then I started hitting shell's cmd line limits. This would be very difficult to guess for each repo. So instead moved the entire thing into the script. 

We still won't be able to build a node action like in `update-action-versions` because we need Go in the running env. Instead I've kept it as a `composite` action.

Tested in https://github.com/smartcontractkit/chainlink-data-streams/actions/runs/9874646427/job/27269498765 and it looks good.

apps/go-mod-validator/action.yaml
```diff
@@ -0,0 +1,81 @@
...
+      uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b # v4.0.3
+      with:
+        node-version: 20
+        node-version-file: package.json
+
+    - name: Run go-mod-validator
+      shell: bash
+      run:
+        GITHUB_TOKEN=${{ inputs.github-token }} node
+        github/apps/go-mod-validator/dist/index.js
```
erikburt, line 69: How is this working if `dist` hasn't been committed?

erikburt (changes requested):

apps/go-mod-validator/action.yaml
```diff
@@ -0,0 +1,85 @@
...
+  checkout-repo:
+    description: "enable git checkout repo"
+    required: false
+    default: "true"
+  checkout-repo-fetch-depth:
+    description: "number of commits to fetch"
+    required: false
+    default: "0"
+  # inputs
+  github_action_ref:
```
manjunathb4461, line 15: We need the code in this ref for us to execute `node dist/index.js` command. Usually this problem doesn't exist as node packages are usually run as `using: node`. But we need to run a composite action as we need to add Go compiler to the environment. 
Another way of solving this problem would be to run `go-mod-validator` as a docker container. That way the image has downloaded all the necessary code we need to run the code. Looks like this is how many other language custom actions are run.
Any other ideas on how to enable us to run `node dist/index.js` in this custom action without having to pass an extra input variable `github_action_ref`?

erikburt: > Any other ideas on how to enable us to run node dist/index.js in this custom action without having to pass an extra input variable github_action_ref?

The easiest way is by making the calling workflow use  `setup-go` before this action, then making this action a nodejs action, and not a composite.

apps/go-mod-validator/src/deps.ts
```diff
@@ -0,0 +1,52 @@
+const { execSync } = require("child_process");
+
+function getGoModFiles(): string[] {
+  try {
+    const output = execSync("find \"$(pwd)\" -type f -name 'go.mod'", {
+      encoding: "utf-8",
+    });
```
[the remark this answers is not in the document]
manjunathb4461: zx looks like something built for writing shell scripts in js style, not sure if I'm right 🤔  

I found only this one [page](https://google.github.io/zx/typescript) about using zx in typescript, where they also mention _Node.js requires additional setup_

They've also mentioned in the documentation it's useful in case we're writing complex shell queries but, we've 2 simple shell queries so I didn't feel we required it.

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+          );
+          validationErr = new Error(
+            `${dependency.Path}@${dependency.Version} not found in the default branch`,
+          );
+        }
+      } catch (err) {
+        console.error(
+          `failed to verify dependency: ${dependency.Path}@${dependency.Version}, with err: ${err}`,
+        );
+        validationErr = err;
```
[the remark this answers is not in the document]
manjunathb4461: Thanks, fixed it

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,148 @@
...
+      `GitHub API request failed: ${response.status} ${response.statusText}`,
+    );
+  }
+
+  const data = await response.json();
+  return data.default_branch;
+}
+
+// TODO: there's no perfect way of finding the branches the commit belongs to in github.
+// Temporarily, parsing the html output of an undocumented API here.
```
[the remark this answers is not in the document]
manjunathb4461: We already have the commit hash by this point of execution, we need to find if that commit belongs to the default branch. The search results you mentioned don't mention the branch names the commit is on.

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,148 @@
...
+      accessToken,
+    );
+  }
+
+  return false;
+}
+
+async function getVersionType(versionString: string) {
+  // matches pseudo versins like v0.0.5-0.20220116011046-fa5810519dcb
+  const pseudoVersionRegex = /-([\d.]*)-([a-f0-9]{12})(?:\+[\w.-]+)?$/;
```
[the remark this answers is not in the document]
manjunathb4461: This regex doesn't match `v0.0.0` in `v0.0.0-20210424002626-9572fd6faeae`, so `[\d.]` is the numbers part. Which is timestamp according to https://go.dev/doc/modules/version-numbers

apps/go-mod-validator/action.yaml
```diff
@@ -0,0 +1,81 @@
+name: go-mod-validator
```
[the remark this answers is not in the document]
manjunathb4461: yes, if the user calls `setup-go` before calling our action then we don't need composite action. I'm happy to make this change, waiting for other's opinion.

apps/go-mod-validator/action.yaml
manjunathb4461: Fixed it, I was testing with one of the previous commits

apps/go-mod-validator/action.yaml
manjunathb4461: That looks like the right solution. Shall I go ahead and implement that or wait for other's opinions?

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,70 @@
+import { validateDependency } from "./github";
+import { getDependenciesMap } from "./deps";
+
+const smartContractKitPrefix = "github.com/smartcontractkit";
+
+async function run() {
+  // Read required env vars
+  const githubToken = process.env.GITHUB_TOKEN || "";
+  if (githubToken == "") {
+    console.error("no GITHUB_TOKEN env variable found");
```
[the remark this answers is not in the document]
manjunathb4461: Now that we're using node github action, I've used `@actions/core` wherever possible. Thanks for this suggestion

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,148 @@
+import fetch from "node-fetch";
+import { parse } from "node-html-parser";
+
+async function getDefaultBranch(
+  repo: string,
+  accessToken: string,
+): Promise<string> {
+  const apiUrl = `https://api.github.com/repos/${repo}`;
```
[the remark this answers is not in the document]
manjunathb4461: Looks good, replace node-fetch with this. Thanks!

apps/go-mod-validator/src/github.ts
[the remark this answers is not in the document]
manjunathb4461: I removed this entire parsing. Another idea I came around was to compare the commit with default branch's head and check the status to verify if the commit is present in the branch or not. I've made the changes to the code, PTAL

apps/go-mod-validator/action.yaml
manjunathb4461: Implemented this change. Also added an example in `README`

apps/go-mod-validator/src/github.ts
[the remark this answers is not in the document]
erikburt: > The search results you mentioned don't mention the branch names the commit is on.

The search results confirm whether the commit is on the default branch or not. You search the repo, and the commit hash, and if there's a search result it's confirmed to be on the default branch.

apps/go-mod-validator/README.md
```diff
@@ -0,0 +1,30 @@
...
+## Example usage
+
+```yaml
+steps:
+  - name: Check out the repository
+    uses: actions/checkout
+  - name: Setup go
+    uses: actions/setup-go
+  - name: go-mod-validation
+    uses: smartcontractkit/.github/apps/go-mod-validator
```
erikburt, line 26:
```suggestion
  - name: Validate go.mod
    uses: smartcontractkit/.github/apps/go-mod-validator@<commit> # go-mod-validator@x.y.z
```

We recommend tagging our monorepo actions with a commit and a commented tag so I typically try to make people who might copy paste, to actually grab a specific version by doing this ([example](https://github.com/smartcontractkit/.github/blob/main/actions/policy-bot-config-validator/README.md)).

erikburt (changes requested):
Looks good!

One last thing is a more comprehensive message for the errors. If a developer who doesn't know anything about this workflow, gets an error, they should be able to understand the problem and the solution from the workflow's output.

apps/go-mod-validator/src/github.ts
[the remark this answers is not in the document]
erikburt: I've now seen your solution and that seems more robust 👍 .

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+
+  if (validationFailedDependencies.length != 0) {
+    core.info(
+      `validation failed for following dependencies:\n${validationFailedDependencies.join("\n")}`,
+    );
+    core.setFailed(
+      `validation failed for ${validationFailedDependencies.length} dependencies`,
+    );
```
erikburt, line 84: I mentioned this in the last review. A developer who updates the go.mod might not know anything about this workflow, and these error messages are not enough to know what went wrong, and how to fix them. 

Some questions that should probably be answered:
- What validation failed?
- Where is the error?
- How can I fix it?

manjunathb4461: My bad, I've fixed it now.
- Errors are printed along with failed dependencies
- Also I've put in how to fix failed dependencies errors in the README and added a message in `core.setFailed` to read to README

Tested here:
- Validation failed - [chainlink-data-streams](https://github.com/smartcontractkit/chainlink-data-streams/actions/runs/9898867493)
- Validation failed - [chainlink](https://github.com/smartcontractkit/chainlink/actions/runs/9898891582)
- Validation successful - [crib](https://github.com/smartcontractkit/crib/actions/runs/9898871252)

apps/go-mod-validator/jest.config.ts
```diff
@@ -0,0 +1,12 @@
+import { Config } from "jest";
```
erikburt, line 1: Whoops I thought my commit deleted this - please remove.

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+        } else {
+          validationFailedDependencies.set(dependencyResult, "unknown");
+        }
+      }
+    }
+  }
+
+  if (validationFailedDependencies.size != 0) {
+    validationFailedDependencies.forEach((val, key) =>
+      core.info(`validation failed for: ${key}, err: ${val}`),
```
erikburt, line 86: Make this a `core.error` so the errors are on the workflow's summary page.

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    );
+    core.setFailed(
+      `validation failed for ${validationFailedDependencies.size} dependencies, pls refer README to fix the errors`,
+    );
```
erikburt, line 90: I think the bare minimum is to have a link to instructions on how to fix. That can be the README or a confluence page.

Ideally the instructions on how to fix are inlined - [example](https://github.com/smartcontractkit/chainlink/actions/runs/9910259951).

erikburt (changes requested):

apps/go-mod-validator/src/parser.ts
HenryNguyen5: Here's a script i whipped up so you dont have to depend on jq. You'll need to convert it to TS but it does work when i tested it on the output of `go list -json -m all > output.txt`

```js
const fs = require("fs");

// We want to parse the following Go data structure into JS objects
// This is the output of `go list -m -json all`
// Example output:
// {
// 	"Path": "github.com/gogo/googleapis",
// 	"Version": "v1.4.1",
// 	"Time": "2021-03-10T14:11:46Z",
// 	"Indirect": true,
// 	"Dir": "/home/parallels/.asdf/installs/golang/1.21.7/packages/pkg/mod/github.com/gogo/googleapis@v1.4.1",
// 	"GoMod": "/home/parallels/.asdf/installs/golang/1.21.7/packages/pkg/mod/cache/download/github.com/gogo/googleapis/@v/v1.4.1.mod",
// 	"GoVersion": "1.12"
// }
//
//  type Module struct {
//         Path       string        // module path
//         Query      string        // version query corresponding to this version
//         Version    string        // module version
//         Versions   []string      // available module versions
//         Replace    *Module       // replaced by this module
//         Time       *time.Time    // time version was created
//         Update     *Module       // available update (with -u)
//         Main       bool          // is this the main module?
//         Indirect   bool          // module is only indirectly needed by main module
//         Dir        string        // directory holding local copy of files, if any
//         GoMod      string        // path to go.mod file describing module, if any
//         GoVersion  string        // go version used in module
//         Retracted  []string      // retraction information, if any (with -retracted or -u)
//         Deprecated string        // deprecation message, if any (with -u)
//         Error      *ModuleError  // error loading module
//         Origin     any           // provenance of module
//         Reuse      bool          // reuse of old module info is safe
//     }

//     type ModuleError struct {
//         Err string // the error itself
//     }

const output = fs.readFileSync("output.txt", "utf8").split("\n");
// store parsed objects
const objects = [];
// keep concatenating lines until the object is complete
let objectString = "";
// keep track of the depth of the object
let objectDepth = 0;
// Process
for (const line of output) {
  if (line.includes("{")) {
    objectDepth++;
  }
  if (line.includes("}")) {
    objectDepth--;
  }

  objectString += line;
  if (objectDepth === 0 && objectString) {
    objects.push(JSON.parse(objectString));
    objectString = "";
  }
}

```

apps/go-mod-validator/src/deps.ts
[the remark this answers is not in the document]
HenryNguyen5: I recommend we use https://github.com/actions/toolkit/tree/main/packages/glob instead

apps/go-mod-validator/src/deps.ts
```diff
@@ -0,0 +1,47 @@
+import { execSync } from "child_process";
+import * as core from "@actions/core";
+
+function getGoModFiles(goModDir: string): string[] {
+  let output = "";
+
+  try {
+    output = execSync(`find ${goModDir} -type f -name 'go.mod'`, {
```
HenryNguyen5, line 8: I'd recommend we use https://github.com/actions/toolkit/tree/main/packages/glob instead

apps/go-mod-validator/src/deps.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  if (output.length == 0) {
+    throw new Error("no go.mod files found");
+  }
+  return output.trim().split("\n");
+}
+
+export function getDependenciesMap(goModDir: string): Map<string, any> {
+  // get all go.mod files
```
HenryNguyen5, line 22: wydt about removing this comment? your func naming is clear here

apps/go-mod-validator/src/deps.ts
```diff
@@ -0,0 +1,47 @@
...
+  const modFilePaths = getGoModFiles(goModDir);
+
+  const dependenciesMap = new Map();
+  modFilePaths.forEach((modFilePath: string) => {
+    core.info(`finding dependencies in ${modFilePath}`);
+
+    // get the dir of go.mod file
+    const modFileDir = modFilePath.slice(0, -6);
```
HenryNguyen5, line 30: it'd be more useful to instead have the comment include an example `modFilePath` so that readers know what you're slicing. im not sure why you're slicing the last ~6 chars

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,114 @@
...
+
+async function getDefaultBranch(
+  owner: string,
+  repo: string,
+  octokitClient: Octokit,
+): Promise<string> {
+  const repoObject = await octokitClient.request(
+    `GET /repos/${owner}/${repo}/`,
+  );
+  return repoObject.data.default_branch;
```
HenryNguyen5, line 11: The octokit client has endpoints as properties, you can see an example here: https://github.com/smartcontractkit/.github/blob/838bbbd24e5c675e0b48844a3fe6593f2bf5eb4a/actions/gha-workflow-validator/src/github.ts#L21

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+async function findCommitInDefaultBranch(
+  owner: string,
+  repo: string,
+  defaultBranch: string,
+  commitSha: string,
+  octokitClient: Octokit,
+) {
+  const compareResult = await octokitClient.request(
```
HenryNguyen5, line 21: Lets use the typed endpoints instead, it's a much more ergonomic experience https://github.com/smartcontractkit/.github/blob/838bbbd24e5c675e0b48844a3fe6593f2bf5eb4a/actions/gha-workflow-validator/src/github.ts#L21

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export async function validateDependency(
+  path: string,
+  version: string,
+  octokitClient: Octokit,
+) {
+  // repo format smartcontractkit/chainlink
+  const repoPathSplit = path.split("/");
+  const owner = repoPathSplit[1];
+  const repo = repoPathSplit[2];
```
HenryNguyen5, line 89:
```suggestion
  const [owner, repo] = path.split("/");

```

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+  if (versionMatch) {
+    return {
+      commitSha: null,
+      tag: versionMatch[1],
+    };
+  }
+}
+
+export async function validateDependency(
```
HenryNguyen5, line 81: This would be a good candidate for a function comment so we know what it does

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,114 @@
...
+  return repoObject.data.default_branch;
+}
+
+async function findCommitInDefaultBranch(
+  owner: string,
+  repo: string,
+  defaultBranch: string,
+  commitSha: string,
+  octokitClient: Octokit,
+) {
```
HenryNguyen5, line 20:
```suggestion
async function isCommitInDefaultBranch(
  owner: string,
  repo: string,
  defaultBranch: string,
  commitSha: string,
  octokitClient: Octokit,
): boolean {
```
naming suggestion to match semantics better

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,114 @@
...
+  );
+}
+
+async function findTagInDefaultBranch(
+  owner: string,
+  repo: string,
+  defaultBranch: string,
+  tag: string,
+  octokitClient: Octokit,
+) {
```
HenryNguyen5, line 36:
```suggestion
async function isTagInDefaultBranch(
  owner: string,
  repo: string,
  defaultBranch: string,
  tag: string,
  octokitClient: Octokit,
): boolean {
```
naming suggestion to match semantics better

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,114 @@
...
+  const repoTags = await octokitClient.request(
+    `GET /repos/${owner}/${repo}/tags`,
+  );
+  for (const repoTag of repoTags.data) {
+    if (repoTag.name != tag) {
+      continue;
+    }
+
+    // check commit is on the default branch
```
HenryNguyen5, line 45:
```suggestion

```
We won't need this with a small change in func naming :)

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,114 @@
...
+  repo: string,
+  defaultBranch: string,
+  tag: string,
+  octokitClient: Octokit,
+) {
+  const repoTags = await octokitClient.request(
+    `GET /repos/${owner}/${repo}/tags`,
+  );
+  for (const repoTag of repoTags.data) {
+    if (repoTag.name != tag) {
```
HenryNguyen5, line 41: When can this happen?

apps/go-mod-validator/src/github.ts
```diff
@@ -0,0 +1,114 @@
...
+      repoTag.commit.sha,
+      octokitClient,
+    );
+  }
+
+  return false;
+}
+
+async function getVersionType(versionString: string) {
+  // matches pseudo versins like v0.0.5-0.20220116011046-fa5810519dcb
```
HenryNguyen5, line 59:
```suggestion
  // matches pseudo versions like v0.0.5-0.20220116011046-fa5810519dcb
```

HenryNguyen5: what makes this a pseudo version? afaik its still valid semver, prerelease format

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+import { getDependenciesMap } from "./deps";
+import * as core from "@actions/core";
+import minimist from "minimist";
+import { Octokit } from "octokit";
+
+const smartContractKitPrefix = "github.com/smartcontractkit";
+
+function getContext() {
+  const argv = minimist(process.argv.slice(2));
+  const { local, tokenEnv, goModDir } = argv;
```
HenryNguyen5, line 11: We should use github action's inputs here rather than parsing args

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    for (let dependency of dependencies) {
+      // handle replace redirectives
+      if (dependency.Replace != undefined) {
+        dependency = dependency.Replace;
+      }
+
+      // `go list -m -json all` also lists the main pacakge, avoid parsing it.
+      // and only validate dependencies belonging to our org
+      if (
+        dependency.Version == undefined ||
```
HenryNguyen5, line 48:
```suggestion
        !dependency.Version ||
```

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  // <failed-dependency, error-string>
+  const validationFailedDependencies: Map<string, string> = new Map();
+
+  // Verify each of the dependencies
+  for (const [file, dependencies] of dependenciesMap.entries()) {
+    for (let dependency of dependencies) {
+      // handle replace redirectives
+      if (dependency.Replace != undefined) {
```
HenryNguyen5, line 41:
```suggestion
      if (dependency.Replace) {
```

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    core.info(`failed to get dependencies, err: ${err}`);
+    core.setFailed(`failed to get dependencies`);
+  }
+
+  // <failed-dependency, error-string>
+  const validationFailedDependencies: Map<string, string> = new Map();
+
+  // Verify each of the dependencies
+  for (const [file, dependencies] of dependenciesMap.entries()) {
+    for (let dependency of dependencies) {
```
HenryNguyen5, line 39:
```suggestion
    for (const dependency of dependencies) {
```

Best to use const whereever possible

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+      if (
+        dependency.Version == undefined ||
+        !dependency.Path.startsWith(smartContractKitPrefix)
+      ) {
+        continue;
+      }
+
+      // prepare dependency result string
+      let dependencyResult = `${dependency.Path}@${dependency.Version}`;
+      if (dependency.Indirect == true) {
```
HenryNguyen5, line 56:
```suggestion
      if (dependency.Indirect) {
```

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+        if (err instanceof Error) {
+          validationFailedDependencies.set(dependencyResult, err.message);
+        } else {
+          validationFailedDependencies.set(dependencyResult, "unknown");
+        }
+      }
+    }
+  }
+
+  if (validationFailedDependencies.size != 0) {
```
HenryNguyen5, line 84: We should really be using annotations for giving developers precise feedback: https://github.com/actions/toolkit/tree/main/packages/core#annotations

With this, they'll know exactly which offending lines to amend

apps/go-mod-validator/src/deps.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  try {
+    output = execSync(`find ${goModDir} -type f -name 'go.mod'`, {
+      encoding: "utf-8",
+    });
+  } catch (error) {
+    throw new Error(`failed to get go.mod files: ${error}`);
+  }
+
+  if (output.length == 0) {
```
HenryNguyen5, line 15:
```suggestion
  if (output.length === 0) {
```

It's best to use strict equality checks

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+
+function getContext() {
+  const argv = minimist(process.argv.slice(2));
+  const { local, tokenEnv, goModDir } = argv;
+
+  const dir = local ? goModDir || `"$(pwd)"` : core.getInput("go-mod-dir");
+
+  const githubToken = local
+    ? process.env[tokenEnv] || ""
+    : core.getInput("github-token");
```
HenryNguyen5, line 17: `getInput` actually parses env vars according to the specified format here: https://github.com/actions/toolkit/blob/main/packages/core/src/core.ts#L128
We can set env vars locally according to the same spec and remove the need for arg parsing that way

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      // validate the dependency
+      try {
+        if (
+          !(await validateDependency(
+            dependency.Path,
+            dependency.Version,
+            octokitClient,
+          ))
+        ) {
```
HenryNguyen5, line 68: how do you feel about splitting this into two lines? wrapping an await statement to perform operations inline is quite hard to read IMO

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+            dependencyResult,
+            "dependency not on default branch",
+          );
+        }
+      } catch (err) {
+        if (err instanceof Error) {
+          validationFailedDependencies.set(dependencyResult, err.message);
+        } else {
+          validationFailedDependencies.set(dependencyResult, "unknown");
+        }
```
HenryNguyen5, line 79:
```suggestion

          validationFailedDependencies.set(dependencyResult, err?.message || "unknown");

```

apps/go-mod-validator/src/index.ts
HenryNguyen5, line 84:
```suggestion
  if (validationFailedDependencies.size > 0) {
```
Having a greater than check here is more meaningful since a negative size is invalid

apps/go-mod-validator/test/deps.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    const result = getDependenciesMap("");
+    expect(result).toEqual(
+      new Map<string, any>([
+        [paths[0], JSON.parse(goListMockOutput1)],
+        [paths[1], JSON.parse(goListMockOutput2)],
+      ]),
+    );
+  });
+
+  it("Fail: should handle no go.mod files found", () => {
```
HenryNguyen5, line 35:
```suggestion
  it("should handle no go.mod files found", () => {
```
nit: we dont list test cases by their expected boolean result, knowing that it handles errors implies that the function fails

apps/go-mod-validator/test/deps.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    const paths: string[] = [];
+
+    (execSync as Mock).mockImplementation((command) => {
```
HenryNguyen5, line 38: We should be able to avoid typecasting here by mocking the module out, take a look at https://github.com/smartcontractkit/.github/blob/main/actions/signed-commits/src/run.test.ts for some inspiration.

apps/go-mod-validator/action.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+inputs:
+  github-token:
+    description:
+      "Github token with read permissions on all smartcontractkit repos"
+    required: true
+    default: ${{ github.token }}
+  go-mod-dir:
+    description: "Common directory where go.mod files are located"
+    required: true
```
HenryNguyen5, line 12:
```suggestion
    default: ${{ github.token }}
  go-mod-dir:
    description: "Common directory where go.mod files are located"
```

I dont think we need to set required since we set a default anyway, unless I am mistaken

apps/go-mod-validator/action.yaml
```diff
@@ -0,0 +1,17 @@
+name: go-mod-validator
+description: "Validates dependencies in go.mod files are tested"
```
HenryNguyen5, line 2:
```suggestion
description: "Validates that the git references contained in dependencies within go.mod files exist in their respective default branches"
```

HenryNguyen5 (changes requested):

apps/go-mod-validator/jest.config.ts
manjunathb4461: Done

apps/go-mod-validator/src/index.ts
manjunathb4461: Done

apps/go-mod-validator/src/deps.ts
manjunathb4461: Yes that looks very good, implemented using that

apps/go-mod-validator/src/deps.ts
manjunathb4461: Done

apps/go-mod-validator/src/github.ts
manjunathb4461: ohh yes, that looks good. Implemented using that

apps/go-mod-validator/src/github.ts
manjunathb4461: done

apps/go-mod-validator/src/index.ts
manjunathb4461: yes, my bad. Fixed it

apps/go-mod-validator/test/deps.test.ts
manjunathb4461: Done

apps/go-mod-validator/action.yaml
manjunathb4461: yes, make senes. Done

apps/go-mod-validator/action.yaml
manjunathb4461: Done

apps/go-mod-validator/src/deps.ts
manjunathb4461: Done

apps/go-mod-validator/src/index.ts
manjunathb4461: Inlining looks good. I've [inlined the instructions](https://github.com/smartcontractkit/chainlink-data-streams/actions/runs/9942495604) similarly and will take a look at showing errors in tabular format in another PR.

apps/go-mod-validator/src/parser.ts
manjunathb4461: Thanks for the script, I've used this one now and removed jq as a dependency

apps/go-mod-validator/src/deps.ts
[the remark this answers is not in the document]
manjunathb4461: glob looks good. Implemented using that

apps/go-mod-validator/src/github.ts
erikburt: This should be `const [_, owner, repo] = path.split("/");` ?

HenryNguyen5: Good catch, it should be 
```ts
const [, owner, repo] = path.split("/");
```

apps/go-mod-validator/src/deps.ts
```diff
@@ -0,0 +1,160 @@
+import { execSync } from "child_process";
+import { dirname } from "path";
+import * as core from "@actions/core";
+import * as glob from "@actions/glob";
+
+/**
+ * Taken from go help list for the command `go list -json -m all`
+ *
+ *  type Module struct {
```
erikburt, line 9: Can we get the link to this?

apps/go-mod-validator/src/deps.ts
```diff
@@ -0,0 +1,160 @@
...
+  // Process
+  for (const l of jsonStr.split("\n")) {
+    if (l.includes("{")) {
+      objDepth++;
+    }
+    if (l.includes("}")) {
+      objDepth--;
+    }
+
+    objStr += l;
```
erikburt, line 98: nit: Can we move this to the first line in the for loop?

apps/go-mod-validator/src/deps.ts
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+ * Taken from go help list for the command `go list -json -m all`
+ *
+ *  type Module struct {
+ *        Path       string        // module path
+ *        Query      string        // version query corresponding to this version
+ *        Version    string        // module version
+ *        Versions   []string      // available module versions
+ *        Replace    *Module       // replaced by this module
```
erikburt, line 14: I wonder if we need to take into account this directive when figuring out which versions are in use.

apps/go-mod-validator/src/deps.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/**
+ * Gets the paths of all go.mod files in the specified directory.
+ *
+ * @param goModDir The directory to search for go.mod files.
+ *
+ * @throws If no go.mod files are found or an error occurs.
+ */
+async function getGoModFiles(goModDir: string): Promise<string[]> {
```
erikburt, line 115: IMO it is not clear that this method is globbing recursively for go.mods. Rename some stuff?

apps/go-mod-validator/action.yaml
```diff
@@ -0,0 +1,17 @@
...
+description:
+  "Validates that the git references contained in dependencies within go.mod
+  files exist in their respective default branches"
+
+inputs:
+  github_token:
+    description:
+      "Github token with read permissions on all smartcontractkit repos"
+    default: ${{ github.token }}
+  go_mod_dir:
```
erikburt, line 11: nit: i think we normally use `kebab-case`

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,79 @@
...
+import { getAllGoModDeps } from "./deps";
+import { FIXING_ERRORS } from "./strings";
+import * as github from "@actions/github";
+import * as core from "@actions/core";
+
+const orgPrefix = "github.com/smartcontractkit";
+
+function getContext() {
+  const goModDir: string = core.getInput("go_mod_dir") || process.cwd();
+  const githubToken: string = core.getInput("github_token") || "";
```
erikburt, line 11: https://github.com/actions/toolkit/blob/main/packages/core/src/core.ts#L118-L120
>  Returns an empty string if the value is not defined.

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,79 @@
+import { isGoModReferencingDefaultBranch } from "./github";
+import { getAllGoModDeps } from "./deps";
+import { FIXING_ERRORS } from "./strings";
+import * as github from "@actions/github";
+import * as core from "@actions/core";
+
+const orgPrefix = "github.com/smartcontractkit";
+
+function getContext() {
+  const goModDir: string = core.getInput("go_mod_dir") || process.cwd();
```
erikburt, line 10: This being defaulted, even though the `action.yml` defaults it to something else.

apps/go-mod-validator/test/deps.test.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as glob from "@actions/glob";
+
+const mockedExecSync = vi.mocked(execSync);
+vi.mock("child_process");
+const mockedGlob = vi.mocked(glob);
+vi.mock("@actions/glob");
+type MockedGlob = MockedObject<Awaited<ReturnType<typeof glob.create>>>;
+
+describe("getDependenciesMap", () => {
+  it("should return a map of <go.mod files: dependencies in json>", async () => {
```
erikburt, line 13: Discussed offline, we should think about switching these to actual `go.mod` files rather than mocks, can be done separately though.

apps/go-mod-validator/test/deps.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    const paths: string[] = [];
+
+    mockedGlob.create.mockResolvedValue({
+      glob: vi.fn().mockResolvedValue(paths),
+    } as MockedGlob);
+
+    await expect(getAllGoModDeps("")).rejects.toThrow("no go.mod files found");
+  });
+
+  it("should handle `find` command failure", async () => {
```
erikburt, line 40: New name? Or does `glob` dispatch to `find` under the hood?

apps/go-mod-validator/test/github.test.ts
```diff
@@ -0,0 +1,114 @@
...
+  rest: {
+    repos: {
+      get: mockGet,
+      compareCommits: mockCompareCommits,
+      listTags: mockListTags,
+    },
+  },
+};
+describe("isGoModReferencingDefaultBranch", () => {
+  describe("pseudo-versions", () => {
```
erikburt, line 22: Discussed offline, we could do this with actual commits and fixtured responses. Can be done in separate PR though.

erikburt (changes requested):

apps/go-mod-validator/src/deps.ts
HenryNguyen5: We do, see goModsToGoModules function

apps/go-mod-validator/test/deps.test.ts
HenryNguyen5: https://smartcontract-it.atlassian.net/jira/software/c/projects/RE/boards/83?selectedIssue=RE-2792

apps/go-mod-validator/test/github.test.ts
HenryNguyen5: https://smartcontract-it.atlassian.net/jira/software/c/projects/RE/boards/83?selectedIssue=RE-2792

apps/go-mod-validator/src/index.ts
```diff
@@ -0,0 +1,44 @@
...
+async function run() {
+  const { goModDir, gh } = getContext();
+
+  const depsToValidate = await getAllGoModDeps(goModDir);
+
+  // <dependency-name, error-string>
+  const errs: Map<string, string> = new Map();
+
+  for (const d of depsToValidate) {
+    const refsDefaultBranch = await isGoModReferencingDefaultBranch(d, gh);
```
erikburt, line 23: isValid?

apps/go-mod-validator/src/deps.ts
```diff
@@ -136,25 +141,54 @@ async function getGoModFiles(goModDir: string): Promise<string[]> {
  * @param rootDir The directory containing the Go modules.
  *
  */
-export async function getAllGoModDeps(rootDir: string): Promise<GoMod[]> {
-  const modFilePaths = await getGoModFiles(rootDir);
+export async function getAllGoModDeps(rootDir: string): Promise<GoModule[]> {
```
erikburt, line 144: This name doesn't make sense anymore now that you're filtering and processing the modules.

erikburt (changes requested):
Previous README.md comments not addressed either.

apps/go-mod-validator/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "description": "Parses 'go.mod' file given as input, and verifies it's dependencies",
+  "engines": {
+    "pnpm": "9",
+    "node": "20"
+  },
+  "main": "out/index.js",
+  "scripts": {
+    "local": "GITHUB_STEP_SUMMARY=/dev/null ts-node src/index.ts",
+    "help": "ts-node --help",
+    "build": "esbuild src/index.ts --bundle --platform=node --target=node20 --tsconfig=./tsconfig.json --outfile=dist/index.js",
```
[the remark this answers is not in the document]
HenryNguyen5: Changed, oddly enough i cant get it to work properly when im running `nx` within the project dir, seems to be tied to the cwd. Got any ideas? Otherwise, running `nx go-mod-validator:build` works fine from the repo root

apps/go-mod-validator/README.md
```diff
@@ -0,0 +1,44 @@
...
+  - name: Validate go.mod
+    uses: smartcontractkit/.github/apps/go-mod-validator@<commit> # go-mod-validator@x.y.z
+```
+
+## Running locally
+
+1. Set github token by setting env variable `INPUT_GITHUB-TOKEN`
+2. Set directory containing `go.mod` files (also any child directories) by
+   setting env var `INPUT_GO-MOD-DIR`, default will be current working directory
+3. Set prefix to filter dependencies by setting env var `INPUT_DEP-PREFIX`
```
erikburt, line 43: `-`s aren't allowed in many shells environment variables. I believe the `core.getInput` method translates all `-` to `_`. So `_`s should work here instead.

erikburt (dismissed):

apps/go-mod-validator/README.md
HenryNguyen5: unfortunately it doesnt, it adds `_` for all whitespace characters https://github.com/actions/toolkit/blob/main/packages/core/src/core.ts#L128

HenryNguyen5 (dismissed):

PR #502, opened 2024-07-08
Add Support For S3 caching to Nix Setup Action

ecPablo:
Usage of action example: https://github.com/smartcontractkit/gauntlet-plus-plus/pull/515

### DIFF
--- a/actions/setup-nix-cache/action.yml
+++ b/actions/setup-nix-cache/action.yml
@@ -94,7 +94,7 @@ runs:
           sudo sed -i "s|^substituters.*|$SUBSTITUTERS_LINE|" /etc/nix/nix.conf
         else
           echo "➕ Appending 'substituters' line to /etc/nix/nix.conf"
-          echo "$SUBSTITUTERS_LINE" | sudo tee -a /etc/nix/nix.conf
+          echo "$SUBSTITUTERS_LINE" >> /etc/nix/nix.conf
         fi
 
         # Replace or append the 'trusted-public-keys' line
@@ -103,7 +103,7 @@ runs:
           sudo sed -i "s|^trusted-public-keys.*|$PUBLIC_KEYS_LINE|" /etc/nix/nix.conf
         else
           echo "➕ Appending 'trusted-public-keys' line to /etc/nix/nix.conf"
-          echo "$PUBLIC_KEYS_LINE" | sudo tee -a /etc/nix/nix.conf
+          echo "$PUBLIC_KEYS_LINE" >> /etc/nix/nix.conf
         fi
 
     # Store private signing key if provided
@@ -124,7 +124,7 @@ runs:
           sudo sed -i "s|^secret-key-files.*|$SECRET_KEY_FILES_LINE|" /etc/nix/nix.conf
         else
           echo "➕ Appending 'secret-key-files' line to /etc/nix/nix.conf"
-          echo "$SECRET_KEY_FILES_LINE" | sudo tee -a /etc/nix/nix.conf
+          echo "$SECRET_KEY_FILES_LINE" >> /etc/nix/nix.conf
         fi
 
     # Set up post-build hook for cache upload if pushing
@@ -166,7 +166,7 @@ runs:
 
         # Set up the post-build hook to use the modified script
         echo "Setting post-build-hook in /etc/nix/nix.conf"
-        echo "post-build-hook = $PERSISTENT_SCRIPT_PATH" | sudo tee -a /etc/nix/nix.conf
+        echo "post-build-hook = $PERSISTENT_SCRIPT_PATH" >> /etc/nix/nix.conf
 
     # Restart Nix daemon for both push and read modes to apply configuration changes
     - name: Restart Nix Daemon
--- a/actions/setup-nix-cache/action.yml
+++ b/actions/setup-nix-cache/action.yml
@@ -1,0 +1,174 @@
+name: setup-nix-cache
+description: "Setup Nix cache with S3 support and optional signing"
+
+inputs:
+  cache-url:
+    description: "Custom Nix cache URL, e.g., for S3 use 's3://<bucket-name>'"
+    required: true
+
+  cache-pubkeys:
+    description: "Comma-separated list of custom Nix cache public keys"
+    required: true
+
+  cache-mode:
+    description:
+      "Specify 'read' for read-only access to the cache or 'push' for uploading
+      build results"
+    required: true
+    default: "read"
+
+  aws-region:
+    description:
+      "AWS region for the S3 bucket (required for accessing S3-backed caches)"
+    required: true
+
+  role-to-assume:
+    description: "AWS role to assume for cache access (required)"
+    required: true
+
+  role-duration-seconds:
+    description: "Duration in seconds for assuming the role (default: 1 hour)"
+    required: false
+    default: 3600
+
+  private-signing-key:
+    description:
+      "Private signing key for Nix cache (only required if pushing to cache)"
+    required: false
+    default: ""
+
+runs:
+  using: composite
+  steps:
+    # Step to configure AWS credentials for Nix cache
+    - name: Configure AWS credentials for Nix cache
+      uses: aws-actions/configure-aws-credentials@050bf7aae915badb82daa8e68fd95c9070706953 # v4.0.2
+      with:
+        role-to-assume: ${{ inputs.role-to-assume }}
+        role-duration-seconds: ${{ inputs.role-duration-seconds }}
+        aws-region: ${{ inputs.aws-region }}
+
+    # Step to set AWS environment variables and symlink the credentials
+    - name: 🛠️ Setup AWS credentials for Nix
+      shell: bash
+      run: |
+        echo "🔑 Setting up AWS credentials for Nix..."
+        aws configure set aws_access_key_id $AWS_ACCESS_KEY_ID
+        aws configure set aws_secret_access_key $AWS_SECRET_ACCESS_KEY
+        aws configure set aws_session_token $AWS_SESSION_TOKEN
+        aws configure set region ${{ inputs.aws-region }}
+
+        ROOT_PATH=""
+        if [[ "${{ runner.os }}" == "Linux" ]]; then
+          ROOT_PATH="/root"
+        elif [[ "${{ runner.os }}" == "macOS" ]]; then
+          ROOT_PATH="/var/root"
+        else
+          echo "⚠️ ${ { runner.os }} not supported!"
+          exit 1
+        fi
+
+        echo "🔗 Symlinking AWS credentials for Nix daemon..."
+        sudo ln -s ~/.aws $ROOT_PATH
+
+    # Step to preprocess public keys
+    - name: Preprocess Public Keys
+      shell: bash
+      run: |
+        PUBLIC_KEYS=$(echo "${{ inputs.cache-pubkeys }}" | tr ',' ' ')
+        echo "PUBLIC_KEYS=$PUBLIC_KEYS" >> $GITHUB_ENV
+
+    # Setup Cache with Nix
+    - name: Setup Nix Cache
+      shell: bash
+      run: |
+        # Define the substitutions and public keys strings
+        SUBSTITUTERS_LINE="substituters = https://cache.nixos.org ${{ inputs.cache-url }}?scheme=https&region=${{ inputs.aws-region }}"
+        PUBLIC_KEYS_LINE="trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= ${{ env.PUBLIC_KEYS }}"
+
+        # Replace or append the 'substituters' line
+        if grep -q "^substituters" /etc/nix/nix.conf; then
+          echo "🔄 Replacing existing 'substituters' line in /etc/nix/nix.conf"
+          sudo sed -i "s|^substituters.*|$SUBSTITUTERS_LINE|" /etc/nix/nix.conf
+        else
+          echo "➕ Appending 'substituters' line to /etc/nix/nix.conf"
+          echo "$SUBSTITUTERS_LINE" | sudo tee -a /etc/nix/nix.conf
+        fi
+
+        # Replace or append the 'trusted-public-keys' line
+        if grep -q "^trusted-public-keys" /etc/nix/nix.conf; then
+          echo "🔄 Replacing existing 'trusted-public-keys' line in /etc/nix/nix.conf"
+          sudo sed -i "s|^trusted-public-keys.*|$PUBLIC_KEYS_LINE|" /etc/nix/nix.conf
+        else
+          echo "➕ Appending 'trusted-public-keys' line to /etc/nix/nix.conf"
+          echo "$PUBLIC_KEYS_LINE" | sudo tee -a /etc/nix/nix.conf
+        fi
+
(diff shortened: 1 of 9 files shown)


actions/setup-nix/README.md
```diff
@@ -1,32 +1,97 @@
...
 ## Configuration
 
+## Inputs
+
+| Name                    | Description                                                                                        | Required | Default     |
+| ----------------------- | -------------------------------------------------------------------------------------------------- | -------- | ----------- |
+| `cache-url`             | Custom Nix cache URL, for example, `'s3://<bucket-name>'` for an S3-backed cache.                  | Yes      | N/A         |
+| `cache-pubkeys`         | Comma-separated list of custom Nix cache public keys.                                              | Yes      | N/A         |
+| `cache-mode`            | Specify either `'read'` for read-only access or `'push'` for uploading build results to the cache. | Yes      | `read`      |
+| `enable-aws`            | Enable AWS credentials for Nix cache (required for S3-backed caches).                              | No       | `false`     |
```
krebernisak, line 16: If this is disabled, then the action will just do the install?

actions/setup-nix/README.md
```diff
@@ -1,32 +1,97 @@
...
+read/write mode.
 
 ## Configuration
 
+## Inputs
+
+| Name                    | Description                                                                                        | Required | Default     |
+| ----------------------- | -------------------------------------------------------------------------------------------------- | -------- | ----------- |
+| `cache-url`             | Custom Nix cache URL, for example, `'s3://<bucket-name>'` for an S3-backed cache.                  | Yes      | N/A         |
+| `cache-pubkeys`         | Comma-separated list of custom Nix cache public keys.                                              | Yes      | N/A         |
```
krebernisak, line 14: This is ok for now, but we should be able to improve on this.

Ideally all trusted pubkeys are published to a single repo, and all other repos just use it from there as configuration, vs. manually specifying their own (duplicates across repos).

actions/setup-nix/action.yml
```diff
@@ -1,107 +1,159 @@
...
+      shell: bash
+      run: |
+        # Preprocess the public keys and store them in a variable
+        PUBLIC_KEYS=$(echo "${{ inputs.cache-pubkeys }}" | tr ',' ' ')
+        echo "Processed Public Keys: $PUBLIC_KEYS"
+
+        # Store the processed public keys in an environment variable for use in later steps
+        echo "PUBLIC_KEYS=$PUBLIC_KEYS" >> $GITHUB_ENV
+
+    - uses: DeterminateSystems/nix-installer-action@da36cb69b1c3247ad7a1f931ebfd954a1105ef14 # v14
```
krebernisak, line 99: v14?

I do feel we could/should extract the install action (third-party) from the cache setup action (our S3 setup). Would be cleaner, and users could configure with custom `install_url` etc.

actions/setup-nix/action.yml
```diff
@@ -1,107 +1,159 @@
...
+      run: |
+        if [ "${{ inputs.cache-mode }}" == "push" ]; then
+          echo "🚀 Enabling push mode: configuring post-build hook for cache upload"
+
+          # Create a modified version of the script with the actual inputs for CACHE_URL and AWS_REGION
+          MODIFIED_SCRIPT_PATH="/tmp/upload-to-cache-modified.sh"
+          cp ${GITHUB_ACTION_PATH}/scripts/upload-to-cache.sh $MODIFIED_SCRIPT_PATH
+
+          # Use sed to replace the placeholders in the script
+          sed -i "s|export CACHE_URL=\$1|export CACHE_URL=${{ inputs.cache-url }}|" $MODIFIED_SCRIPT_PATH
```
krebernisak, line 119: Hmm, do we really need to modify the script.

Could we use bash args in the `post-build-hook` or set an env variables before running the script?

krebernisak (dismissed):

actions/setup-nix/action.yml
ecPablo: unfortunately the post build hook does not support passing params. you can only pass the script path. Env variables could be an option but using the composite actions we are not able to pass env variables here. We would need to rely on the user knowing the env variable is setup before calling the action.

actions/setup-nix/action.yml
ecPablo: One issue I see here is that the cache setup paths might change depending on the gh action used to install nix. For example I'm not 100% sure if all installer place configs under `/etc/nix/nix.conf`.It might make it a bit harder to support all possible nix installers. But I agree it would make the params a bit cleaner in general.

actions/setup-nix/README.md
ecPablo: Yeah make sense. Mitchell is also provisioning them as [Org Wide Variables ](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#creating-configuration-variables-for-an-organization) so it's another option too.

actions/setup-nix/action.yml
krebernisak: As an idea, there is an $NIX_CONF_DIR env var that should be available.

https://nix.dev/manual/nix/2.24/command-ref/conf-file.html

actions/setup-nix/action.yml
krebernisak: Couldn't the env var be set up in the action itself?

The input params to the action are both the AWS_REGION and CACHE_URL. Not sure if we can map this input to env vars for the action/job.

ecPablo: I just tried this https://github.com/smartcontractkit/.github/pull/502/commits/e1678082bd30b8ef19031eb6abd3b8c8a739ee3d , but unfortunately it seems that after setting the env var value in the composite action and the composite action returning to the main workflow, the value gets resetted and subsequent uploads on the workflow fail: https://github.com/smartcontractkit/gauntlet-plus-plus/actions/runs/11145685297/job/30975857353?pr=515#step:4:2997

I will keep the original approach althoguh it's a bit more ugly.

actions/setup-nix/README.md
ecPablo: I was able to separate into 2 actions so now it should be a lot simpler https://github.com/smartcontractkit/.github/pull/502/commits/f903c07a199c1b28a8256c7e10b4c2d2a45f6db7

actions/setup-nix/action.yml
ecPablo: I was able to separate the install / cache action now https://github.com/smartcontractkit/.github/pull/502/commits/f903c07a199c1b28a8256c7e10b4c2d2a45f6db7

For now im defaulting to `/etc/nix/nix.conf` as the `NIX_CONF_DIR` appears to be empty even after installing Nix. Let me know your thoughts with this new separations!

actions/setup-nix-cache/scripts/upload-to-cache.sh
```diff
@@ -0,0 +1,26 @@
...
+export AWS_REGION=$2
+
+# Add your upload logic here, using the environment variables
+echo "Uploading to cache at $CACHE_URL in region $AWS_REGION..."
+# Update PATH
+export PATH=/home/runner/.nix-profile/bin/nix:$PATH
+
+# Log and upload the paths to the Nix cache
+echo "Uploading paths $OUT_PATHS to $CACHE_URL with AWS region $AWS_REGION"
+exec nix copy --to "$CACHE_URL?scheme=https&region=$AWS_REGION" $OUT_PATHS
```
krebernisak, line 26: Is the signing happening automatically during build or during `nix copy`, with a configured secret key?

krebernisak (approved):
Thx for splitting up the actions 👍  LGTM!

actions/setup-nix-cache/scripts/upload-to-cache.sh
ecPablo: If I'm not wrong, the derivations get signed as they are built. So here we are just uploading already signed artifacts.

actions/setup-nix-cache/action.yml
```diff
@@ -0,0 +1,174 @@
...
+
+    # Step to set AWS environment variables and symlink the credentials
+    - name: 🛠️ Setup AWS credentials for Nix
+      shell: bash
+      run: |
+        echo "🔑 Setting up AWS credentials for Nix..."
+        aws configure set aws_access_key_id $AWS_ACCESS_KEY_ID
+        aws configure set aws_secret_access_key $AWS_SECRET_ACCESS_KEY
+        aws configure set aws_session_token $AWS_SESSION_TOKEN
+        aws configure set region ${{ inputs.aws-region }}
```
chainchad, line 59: We have to do these because Nix can't work with the secrets available to it as env vars? If so, I think it'd be worth documenting this limitation and possibly looking at an alternative (if we haven't yet). Normally best practice would dictate not having to write these to disk (the AWS config file), which these commands do.

actions/setup-nix-cache/action.yml
```diff
@@ -0,0 +1,174 @@
...
+
+        ROOT_PATH=""
+        if [[ "${{ runner.os }}" == "Linux" ]]; then
+          ROOT_PATH="/root"
+        elif [[ "${{ runner.os }}" == "macOS" ]]; then
+          ROOT_PATH="/var/root"
+        else
+          echo "⚠️ ${ { runner.os }} not supported!"
```
chainchad, line 67:
```suggestion
          echo "⚠️ ${{ runner.os }} not supported!"
```

actions/setup-nix-cache/action.yml
```diff
@@ -0,0 +1,174 @@
...
+      shell: bash
+      run: |
+        # Define the substitutions and public keys strings
+        SUBSTITUTERS_LINE="substituters = https://cache.nixos.org ${{ inputs.cache-url }}?scheme=https&region=${{ inputs.aws-region }}"
+        PUBLIC_KEYS_LINE="trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= ${{ env.PUBLIC_KEYS }}"
+
+        # Replace or append the 'substituters' line
+        if grep -q "^substituters" /etc/nix/nix.conf; then
+          echo "🔄 Replacing existing 'substituters' line in /etc/nix/nix.conf"
+          sudo sed -i "s|^substituters.*|$SUBSTITUTERS_LINE|" /etc/nix/nix.conf
```
chainchad, line 92: Will this command show the bucket URI in the logs?

actions/setup-nix-cache/action.yml
```diff
@@ -0,0 +1,174 @@
...
+        else
+          echo "➕ Appending 'substituters' line to /etc/nix/nix.conf"
+          echo "$SUBSTITUTERS_LINE" | sudo tee -a /etc/nix/nix.conf
```
chainchad, line 95: Will this command show the bucket URI in the logs?

To be safe, maybe just use `>>` to append to this file without showing to stdout like `tee -a`?

actions/setup-nix-cache/action.yml
```diff
@@ -0,0 +1,174 @@
...
+        PUBLIC_KEYS=$(echo "${{ inputs.cache-pubkeys }}" | tr ',' ' ')
+        echo "PUBLIC_KEYS=$PUBLIC_KEYS" >> $GITHUB_ENV
+
+    # Setup Cache with Nix
+    - name: Setup Nix Cache
+      shell: bash
+      run: |
+        # Define the substitutions and public keys strings
+        SUBSTITUTERS_LINE="substituters = https://cache.nixos.org ${{ inputs.cache-url }}?scheme=https&region=${{ inputs.aws-region }}"
+        PUBLIC_KEYS_LINE="trusted-public-keys = cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY= ${{ env.PUBLIC_KEYS }}"
```
chainchad, line 87: Should there be a space here?

`Y= $`

actions/setup-nix-cache/action.yml
```diff
@@ -0,0 +1,174 @@
...
+          echo "$SUBSTITUTERS_LINE" | sudo tee -a /etc/nix/nix.conf
+        fi
+
+        # Replace or append the 'trusted-public-keys' line
+        if grep -q "^trusted-public-keys" /etc/nix/nix.conf; then
+          echo "🔄 Replacing existing 'trusted-public-keys' line in /etc/nix/nix.conf"
+          sudo sed -i "s|^trusted-public-keys.*|$PUBLIC_KEYS_LINE|" /etc/nix/nix.conf
+        else
+          echo "➕ Appending 'trusted-public-keys' line to /etc/nix/nix.conf"
+          echo "$PUBLIC_KEYS_LINE" | sudo tee -a /etc/nix/nix.conf
```
chainchad, line 104: Will this command show the bucket URI in the logs?

To be safe, maybe just use `>>` to append to this file without showing to stdout like `tee -a`?

actions/setup-nix-cache/action.yml
chainchad: I _think_ it will mask it, but I would test this on a calling workflow by passing in a non-secret to confirm.

actions/setup-nix-cache/action.yml
```diff
@@ -0,0 +1,174 @@
...
+
+        # Ensure the modified script has the right permissions
+        chmod 755 $MODIFIED_SCRIPT_PATH
+
+        # Copy the modified script to a persistent location
+        PERSISTENT_SCRIPT_PATH="/etc/nix/upload-to-cache.sh"
+        sudo cp $MODIFIED_SCRIPT_PATH $PERSISTENT_SCRIPT_PATH
+
+        # Check if the modified script exists and has correct permissions
+        if [ -f "$PERSISTENT_SCRIPT_PATH" ]; then
```
chainchad, line 157:
```suggestion
        if [[ -f "$PERSISTENT_SCRIPT_PATH" ]]; then
```

May as well since we're using bash and to be consistent.

actions/setup-nix-cache/action.yml
ecPablo: Yes unfortunately there is not alternative to this we need to set this directly on the aws environment for nix to be able to authenticate correctly. Documented it here https://github.com/smartcontractkit/.github/pull/502/commits/1dd10b0adb2db32a05311b6274cb62208bbe5e5b

actions/setup-nix-cache/action.yml
ecPablo: Removed any attempts to show the uri on logs here https://github.com/smartcontractkit/.github/pull/502/commits/c210d428bb50ccbabf2732e541cd7eb9515e0e7c

actions/setup-nix-cache/action.yml
ecPablo: Removed all instances of `tee -a` here to be safe https://github.com/smartcontractkit/.github/pull/502/commits/c210d428bb50ccbabf2732e541cd7eb9515e0e7c

actions/setup-nix-cache/action.yml
ecPablo: Yes the space is correct, each publick key is separated by a space. The first one is the default nixOS packages public key which ends in `=` after that we have the space and then whatever the other public keys are

actions/setup-nix-cache/action.yml
ecPablo: removed here too thanks! https://github.com/smartcontractkit/.github/commit/c210d428bb50ccbabf2732e541cd7eb9515e0e7c

chainchad (approved):

PR #509, opened 2024-07-16
RE-1920 implement revoke gati token

manjunathb4461:
Github doesn’t allow explicit ordering of steps, meaning we cannot revoke the token within our action `setup-github-token`. Instead we will have to call a step separately at the end of all steps in the job to revoke the token.

Since it is going to be one single command, creating a separate action seems redundant. I've implemented a change to the existing `setup-github-token` to also allow user's to revoke the token and add an additional input revoke which when set will revoke the token. We can even rename this action to  `github-token-issuer` to avoid the confusion.

Tested here - https://github.com/smartcontractkit/mercury-pipeline/pull/1032/checks

### DIFF
--- a/.changeset/few-mails-act.md
+++ b/.changeset/few-mails-act.md
@@ -1,0 +1,5 @@
+---
+"setup-github-token": minor
+---
+
+Add support for revoking the token at the end of the job
--- a/actions/setup-github-token/README.md
+++ b/actions/setup-github-token/README.md
@@ -1,22 +1,14 @@
 # setup-github-token action
 
 ## Complete example usage
-jobs:
-  release:
-    name: Release
-    runs-on: ubuntu-latest
-    permissions:
-      id-token: write
-      contents: read
-    steps:
-      - name: Setup GitHub Token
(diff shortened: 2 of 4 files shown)


HenryNguyen5 (changes requested):
We should use the `post` lifecycle of the `setup-github-token` action instead. It won't require any changes on the workflow author side.

manjunathb4461:
> We should use the `post` lifecycle of the `setup-github-token` action instead. It won't require any changes on the workflow author side.

The post option is only available for actions using node. When running using docker there’s similar option for post-entrypoint. But composite actions don’t have that option according to their [schema](https://json.schemastore.org/github-action.json), which our [setup-github-token](https://github.com/smartcontractkit/.github/blob/main/actions/setup-github-token/action.yml#L35) action uses. I’m assuming because composite actions have the flexibility of defining more than one step to run and don’t need such options.

momentmaker:
closing this with what Henry had suggested above:
https://github.com/smartcontractkit/.github/pull/509#pullrequestreview-2181847087

PR #516, opened 2024-07-24
RE 2792 write integration tests for go mod validator script

HenryNguyen5:
- **Add e2e test data**
- **Move entrypoint to diff file to make e2e testing easy**
- **Better logging for glob func**
- **Map before filter to avoid missing results**
- **Parallelize github requests**
- **Handle GH API throttling**
- **Have run() return the summary as a string too**
- **Add e2e tests**


chainchad (approved):

momentmaker (approved):

apps/go-mod-validator/src/go-mod-validator.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const validating = depsToValidate.map(async (d) => {
+    const isValid = await isGoModReferencingDefaultBranch(d, gh);
+
+    if (!isValid) {
+      errs.set(d.name, "dependency not on default branch");
+    }
+  });
+
+  await Promise.all(validating);
```
erikburt, line 57: The for loop was far more readable. I understand this parallelizes the loop, and then you wait for the promises to settle, but this is not performance critical code.

I think readability should be preferred here.

apps/go-mod-validator/test/e2e.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  process.env["INPUT_DEP-PREFIX"] = "github.com/smartcontractkit";
+  process.env["GITHUB_STEP_SUMMARY"] = "/dev/null";
+}
+
+describe("e2e tests", () => {
+  it("chainlink - should match snapshot", { timeout: 20_000 }, async () => {
+    setup("chainlink");
+    const summary = await run();
+
+    expect(summary).toMatchInlineSnapshot(`
```
erikburt, line 19: I wouldn't fully match the contents. Any update to the summary text requires an update to the test. This is more annoying than helpful.

Maybe `run` should return an object that has some boolean indicating whether validation passed, and you can use that for validation rather than strings that may change.

apps/go-mod-validator/test/e2e.test.ts
```diff
@@ -0,0 +1,81 @@
...
+const testDataDir = "./test/data";
+
+function setup(repoName: string) {
+  process.env["INPUT_GO-MOD-DIR"] = join(testDataDir, repoName);
+  process.env["INPUT_GITHUB-TOKEN"] = process.env.GITHUB_TOKEN;
+  process.env["INPUT_DEP-PREFIX"] = "github.com/smartcontractkit";
+  process.env["GITHUB_STEP_SUMMARY"] = "/dev/null";
+}
+
+describe("e2e tests", () => {
```
erikburt, line 14: These tests are only validating the binary validation result. Should there be better validation for the individual errors?

apps/go-mod-validator/test/e2e.test.ts
HenryNguyen5: This text is going to be quite static once i make use of the annotation API. It's easy to update snapshots as its automated via the `-u` flag.

apps/go-mod-validator/test/e2e.test.ts
HenryNguyen5: I tried to keep it pretty KISS as they're e2e tests, we can add more complex ones later on if individual unit tests dont suffice

apps/go-mod-validator/src/go-mod-validator.ts
HenryNguyen5: This shortens run-time from 30+s to <5s. It's a large enough difference to keep around. It was especially apparent during e2e testing.

PR #517, opened 2024-07-26
fix: split signed commits gql payload

erikburt:
### Changes

- Split GraphQL mutations by package

### Testing

Tested locally with:

```
DEBUG=true INPUT_SETUPGITUSER=false INPUT_CWD="/Users/erik/Documents
/repos/external-adapters-js" GITHUB_TOKEN=<personal FGAT> INPUT_PRDRAF
T=true GITHUB_REPOSITORY="erikburt/external-adapters-js" GITHUB_REF="r
efs/heads/main" GITHUB_SHA="6bbb8354ff62158d80d073c0611a4566e4d5de32" 
node actions/signed-commits/dist/index.js
```

### TODO

- Better debug logging
- Commit file deletions (only additions done)
- Unit tests


momentmaker (approved):
nice :)

actions/signed-commits/src/git/github-git/index.ts
```diff
@@ -21,30 +23,56 @@ export async function commitAll(
...
+      },
+      message: {
+        headline: commit.message,
+        body: "",
+      },
+      expectedHeadOid,
+      fileChanges: commit.fileChanges,
+    };
+    const response = await createCommitOnBranch(client, input);
+    expectedHeadOid = getRemoteHeadOidFromResponseOrThrow(response);
```
HenryNguyen5, line 60: If this errors mid-way, is it possible that we get stuck in a bad state that would require manual intervention vs being able to re-run the action?

The current edge case I can think of is:

- A transient network error occurs between 2 commits, one deletion commit followed by one addition commit 
- If the action is re-run, we may miss updating changelogs / package.json's since the changesets no longer exist

HenryNguyen5: This commit applies to this entire loop now that I think of it, we need to make sure all changes are applied in serial order and can be recovered from in case of error

HenryNguyen5 (changes requested):

actions/signed-commits/src/git/github-git/index.ts
erikburt: I believe rerunning the workflow would solve the problem wouldn’t it?

HenryNguyen5: If the workflow is re-ran after the addition commit was already pushed but the deletion commit failed, it sounds like it'll create a commit conflict or miss updating changelogs / package.jsons

erikburt: https://github.com/smartcontractkit/.github/blob/da1c0099f3bab236660ccc10328a29e06950c69f/actions/signed-commits/src/run.ts#L341-L360

HenryNguyen5 (approved):

HenryNguyen5 (approved):

PR #523, opened 2024-07-30
Fix payload template for slack-notify-git-ref action

chainchad:


actions/slack-notify-git-ref/action.yml
```diff
@@ -113,7 +121,11 @@ runs:
         fi
 
         cat payload_temp.json
```
momentmaker, line 123: do we need this anymore when we have the following?

actions/slack-notify-git-ref/payload.json
```diff
@@ -1,27 +1,27 @@
 {
-  "text": "New git ref for ${{ env.GITHUB_REPO }}",
+  "text": "New git ref for <https://github.com/${GITHUB_REPOSITORY}|${GITHUB_REPOSITORY}>",
```
momentmaker, line 2: hmm why is GITHUB_REPOSITORY referenced twice?

chainchad: The second one is the anchor text.

momentmaker (approved):

PR #539, opened 2024-08-05
Expose devspace namespace

scheibinger:
* Expose devspace namespace so it can be read by other steps in the workflow
* rename to lower case for consistency

### DIFF
--- a/actions/setup-crib-environment/action.yml
+++ b/actions/setup-crib-environment/action.yml
@@ -55,6 +55,8 @@ inputs:
     default: core
     description: Product name
     required: false
+outputs:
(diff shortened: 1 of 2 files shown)


actions/setup-crib-environment/action.yml
```diff
@@ -55,6 +55,8 @@ inputs:
     default: core
     description: Product name
     required: false
+outputs:
```
njegosrailic, line 58: It looks good can you please add a new release using pnpm?

njegosrailic (approved):

njegosrailic (approved):

PR #545, opened 2024-08-06
Run teardown in the post step

scheibinger:
Execute teardown in the post step to allow users of the workflow to run tests against provisioned env.


scheibinger:
post step doesn't work in regular actions, it only work in javascript based actions

PR #551, opened 2024-08-07
Parameterize devspace profiles

scheibinger:
Provide a way to specify which devspace profiles should be applied during `devspace deploy`

required for: https://smartcontract-it.atlassian.net/browse/CRIB-291


actions/setup-crib-environment/scripts/parse_profiles.sh
```diff
@@ -0,0 +1,26 @@
...
+  for profile in "${profiles_array[@]}"; do
+    result+="-p $profile "
+  done
+
+  # Remove trailing space
+  result=$(echo $result | sed 's/ *$//')
+
+  # Output the result
+  echo "Profiles args: $result"
+  echo "::set-output name=profile_args::$result"
```
momentmaker, line 22: I think `::set-output` is deprecated?

now it's `"echo profile_args=$result" >> GITHUB_OUTPUT`

actions/setup-crib-environment/scripts/parse_profiles.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
```
chainchad, line 2:
```suggestion
#!/usr/bin/env bash
set -euo pipefail
```

actions/setup-crib-environment/scripts/parse_profiles.sh
```diff
@@ -0,0 +1,26 @@
...
+profiles=$PROFILES
+
+if [ -n "$profiles" ]; then
```
chainchad, line 5: Nit, but since we're using bash anyway, I'd use double brackets where possible.

actions/setup-crib-environment/scripts/parse_profiles.sh
```diff
@@ -0,0 +1,26 @@
...
+
+  # Remove trailing space
+  result=$(echo $result | sed 's/ *$//')
+
+  # Output the result
+  echo "Profiles args: $result"
+  echo "::set-output name=profile_args::$result"
+else
+  echo "No profiles supplied"
+  echo "::set-output name=profile_args::"
```
chainchad, line 25: GitHub Actions is trying to deprecate setting outputs like this (see: [this example instead](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/defining-outputs-for-jobs#example-defining-outputs-for-a-job))
```suggestion
  echo "profile_args=$result" | tee -a "$GITHUB_OUTPUT"
else
  echo "No profiles supplied"
  echo "profile_args=" | tee -a "$GITHUB_OUTPUT"
```

Not sure if `echo "profile_args=" | tee -a "$GITHUB_OUTPUT"` is valid -- if not, we can consider just not outputting anything (probably better).

chainchad (changes requested):

actions/setup-crib-environment/scripts/parse_profiles.sh
chainchad: Beat me

scheibinger: fixed

actions/setup-crib-environment/scripts/parse_profiles.sh
scheibinger: fixed

lefthook.yml
```diff
@@ -3,5 +3,5 @@ pre-commit:
   commands:
     prettier:
       description: "Check for prettier formatting"
-      run: pnpm run prettier:check {staged_files}
```
scheibinger, line 6: `{staged_files}` is cool, but I'm not able to exclude .sh scripts from that pattern, as prettier doesn't work for shell files

actions/setup-crib-environment/action.yml
```diff
@@ -152,7 +159,19 @@ runs:
...
+          IFS=',' read -r -a profiles_array <<< "$PROFILES"
+
+          for profile in "${profiles_array[@]}"; do
+            profile_args_array+=("-p")
+            profile_args_array+=("$profile")
+          done
+          echo "running devspace with profile args: ${profile_args_array[@]}"
+        fi
+
+        nix develop -c devspace deploy --skip-build -o=${{ inputs.image-tag }} "${profile_args_array[@]}"
```
scheibinger, line 174: Initially I had it in the separate script, but it didn't work when passing as string. It only works if we pass this as an array

njegosrailic (commented):
It looks good please add a changeset.

njegosrailic (approved):

PR #575, opened 2024-08-28
RE-2934 RE-2933 RE-2943 RE-2944

HenryNguyen5:
- **Update go.mod + go.sum for integration tests**
- **Add simple tag parsing test**
- **Use ref lookup to avoid tag listing pagination, expose commit sha**
- **Update go.mods for integration-tests/load**
- **Add more context to validation failures**
- **Add changeset**


apps/go-mod-validator/src/github.ts
```diff
@@ -23,35 +23,62 @@ async function isCommitInDefaultBranch(
...
   mod: GoModuleWithTag,
-): Promise<boolean> {
-  const { data: repoTags } = await gh.rest.repos.listTags({
-    owner: mod.owner,
-    repo: mod.repo,
-  });
+): Promise<{
+  isInDefault: boolean;
+  commitSha: string;
+}> {
```
erikburt, line 40: Use the interface type here instead?

apps/go-mod-validator/src/github.ts
```diff
@@ -23,35 +23,62 @@ async function isCommitInDefaultBranch(
...
-      continue;
+    // lightweight tags show up as "commit" type, annotated tags show up as "tag" type
+    const isAnnotatedTag = tag.data.object.type === "tag";
+    if (isAnnotatedTag) {
+      const getTagResp = await gh.rest.git.getTag({
+        owner: mod.owner,
+        repo: mod.repo,
+        tag_sha: tag.data.object.sha,
+      });
+      commitSha = getTagResp.data.object.sha;
```
erikburt, line 57: So it didn't support annotated tags before?

apps/go-mod-validator/src/github.ts
```diff
@@ -23,35 +23,62 @@ async function isCommitInDefaultBranch(
...
+      commitSha = tag.data.object.sha;
     }
 
     return isCommitInDefaultBranch(gh, defaultBranch, {
       ...mod,
-      commitSha: repoTag.commit.sha,
+      commitSha,
     });
+  } catch (e) {
+    core.error(`Error checking tag in default branch: ${e}`);
```
erikburt, line 67: Include the owner, repo, and sha?

erikburt (changes requested):
Small comments but you have some test failures as well.

apps/go-mod-validator/src/github.ts
HenryNguyen5: It did, `git.getTag` only supports annotated tags but `repo.listTags` supports all tag types

apps/go-mod-validator/src/github.ts
HenryNguyen5: Good catch

HenryNguyen5: I made it such that there's a warning annotation that includes this context now

apps/go-mod-validator/src/github.ts
erikburt: Then why is this logic required for annotated tags?

HenryNguyen5: we're not using `repo.listTags` anymore because its an inefficient API call. The only way to retrieve a single tag is to use the git database API instead which requires differentiation between the two tag types

apps/go-mod-validator/src/github.ts
```diff
@@ -64,19 +61,39 @@ async function isTagInDefaultBranch(
...
+  /**
+   * Additional information when "isInDefault" is "unknown"
+   */
+  reason: string;
+  commitSha: string;
+}
+
+export type GoModDefaultBranchLookupResult =
+  | GoModDefaultBranchKnownLookupResult
+  | GoModDefaultBranchUnknownResult;
```
erikburt, line 95: Would an enum for `isInDefault` be better than this union type?

apps/go-mod-validator/src/go-mod-validator.ts
```diff
@@ -42,18 +42,18 @@ export async function run(): Promise<string> {
   const { goModDir, gh, depPrefix } = getContext();
 
   const depsToValidate = await getDeps(goModDir, depPrefix);
-
-  const errs: Map<BaseGoModule, string> = new Map();
+  interface Invalidation {
+    type: "error" | "warning";
+    msg: string;
+  }
```
erikburt, line 48: Why inline the interface?

apps/go-mod-validator/test/github.test.ts
```diff
@@ -101,47 +115,62 @@ describe("isGoModReferencingDefaultBranch", () => {
     };
 
     it("should check if the commit is present on the default branch", async () => {
```
erikburt, line 117: You seem to validating calls to mocks in some tests and not in others. Why?

package.json
```diff
@@ -8,6 +8,8 @@
   "scripts": {
     "prettier:check": "prettier --check \"**/*.{json,md,js,ts,yml,yaml}\" --ignore-path=.prettierignore",
     "prettier:write": "prettier --write \"**/*.{json,md,js,ts,yml,yaml}\" --ignore-path=.prettierignore",
```
erikburt, line 10: These are redundant now?

apps/go-mod-validator/src/go-mod-validator.ts
HenryNguyen5: We dont need to expose it anywhere yet

package.json
HenryNguyen5: They're used in CI

apps/go-mod-validator/src/github.ts
HenryNguyen5: With an enum, we'll still have a union type if we want to perform a discriminated union check to differentiate between an `unknown` state vs a `known` one. I initially was going to use an enum but TBH its overkill ATM.

apps/go-mod-validator/test/github.test.ts
HenryNguyen5: The ones where I intentionally validate calls is b/c I cannot infer the lack of a call based on the return result of the function under test. I need to use `not.toHaveBeenCalled` in those cases

apps/go-mod-validator/src/go-mod-validator.ts
erikburt: Maybe this is personal preference but I'd rather see the type in-lined into the Map (as it's only used once), or the interface outside the method.

Thoughts?

apps/go-mod-validator/test/github.test.ts
erikburt: so `mockResolvedValueOnce` will throw/cause errors if something has been called more than once, *and* if it wasn't called at all?

apps/go-mod-validator/src/go-mod-validator.ts
HenryNguyen5: I figured that this interface is going to be expanded upon in the future during a refactoring, its a core data structure. I'd rather keep it as is

apps/go-mod-validator/test/github.test.ts
HenryNguyen5: no, im saying that i dont use any `not.toHaveBeenCalled` calls within the `versions` test block. I do see where I can remove a validation call now since we return an object that contains a test value rather than just a boolean. Making this change

erikburt (approved):

PR #578, opened 2024-08-28
Create license-checker action

chainchad:


actions/license-checker/action.yml
```diff
@@ -0,0 +1,66 @@
...
+  gc-org-id:
+    description:
+      "grafana org/tenant id - required if metrics-job-name is passed"
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - name: "Install license-finder"
+      run: gem install license_finder -v ${{ inputs.license-finder-version }}
```
momentmaker, line 49: do you get `gem` for free with gha runner?
or do you need something like this?
https://github.com/ruby/setup-ruby

chainchad: I think it's available but I have to use sudo to install to the system-wide gem dir, which may not be ideal?

momentmaker: cool as long as it works :)

momentmaker (dismissed):

momentmaker (dismissed):

momentmaker (approved):

PR #585, opened 2024-08-30
feat: refactor gha-workflow-validator validation

erikburt:
### Changes

- Refactor gha-workflow-validator to better support new types of validation
- Add support for running against all content, not just diff content
- Added local testing script

### Notes

Precursor for https://smartcontract-it.atlassian.net/browse/RE-2606

This refactors how things are organized so we can realistically add validations to any kinds of lines. The typing and logic is mostly sound. Hoping to clean most of this up in a subsequent PR which utilizes the refactor to validate runner types.


actions/gha-workflow-validator/src/utils.ts
```diff
@@ -1,27 +1,72 @@
 import * as core from "@actions/core";
+import * as glob from "@actions/glob";
+import { sep } from "path";
+import { readFile } from "fs/promises";
```
momentmaker, line 4: is there a reason to use `readFile` instead of `readFileSync`?

i think the latter is faster?

https://gist.github.com/FauxFaux/079dc1c696cd60c8120a5ed9b18942c3

momentmaker (approved):
🚀

actions/gha-workflow-validator/src/utils.ts
erikburt: Will update in #593

PR #597, opened 2024-09-05
Add PR based validation to go-mod-validator

HenryNguyen5:
- **Add pseudo-version documentation**
- **Add diffing supp on PR**
- **Refactor e2e tests to not require network connection**

Now, when we're in a pull request context, we only run validation against dependencies that are part of the PR diff. We also do not check tagged dependencies in PR mode, only in non-PR mode.

A significant amount of the changes involved in this PR revolves around refactoring the e2e tests so that they are deterministic and maintainable.

[A script was added](https://github.com/smartcontractkit/.github/blob/RE-2926-as-a-developer-id-like-to-have-go-mod-validation-working-for-chainlink/apps/go-mod-validator/test/__helpers__/generate-gomod-fixtures.ts) which allows us to point to a local git directory along with a commit to use. It grabs all of the `go.mod` and `go.sum` files in that dir and updates our fixtures, namespaced by that commit. This lets us easily test against different commits when we discover bugs, something that was manually done before, and was quite tedious.

Along with the change to easily update fixutres, we're now using nock to intercept our calls to github, speeding up tests and removing the need to have github auth in our CI to run them too.

Resolves child tickets: RE-2955 RE-2946


apps/go-mod-validator/test/e2e.test.ts
```diff
@@ -22,186 +34,147 @@ vi.mock("@actions/core", async (importOriginal: any) => ({
   },
 }));
 
-function setup(repoName: string) {
-  process.env["INPUT_GO-MOD-DIR"] = join("./test/data", repoName);
+const nock = getNock("lockdown");
```
erikburt, line 37: Why passing "lockdown"?

erikburt (approved):

.github/workflows/pull-request-main.yml
```diff
@@ -107,30 +107,13 @@ jobs:
...
     runs-on: ubuntu-latest
-    permissions:
-      id-token: write
-      contents: read
-      actions: read
     steps:
       - name: Checkout repo (needed to reference local action)
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
 
-      - name: Get GATI read-only test token
```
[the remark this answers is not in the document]
HenryNguyen5: yeah

apps/go-mod-validator/test/e2e.test.ts
HenryNguyen5: Oh just just to make it explicit i guess, i didnt realize its the default

erikburt: Passing it explicitly doesn't let you update the fixtures through the ENV.

PR #608, opened 2024-09-18
[TT-1693] reusable solidity artifacts workflow

Tofel:
This PR introduces a reusable Solidity Review Artifact generating pipeline that currently was stored in `chainlink` repository.

I have made it more universal, so that it's possible to use with repositories that use different project structure (`chainlink` repo is very specific in that regard).

I have tested it with typical Foundry layout and it works like a charm. Here's a working example: https://github.com/smartcontractkit/solidity-review-artifacts-example that also describes how to use this pipeline in any repository.

Apart from the workflow itself I had to move a couple of scripts and actions:
* bash scripts used by the pipeline
* actions used to setup the pipeline 
* TypeScript code that's responsible for interaction with Jira

You can find some sample executions here: https://github.com/smartcontractkit/solidity-review-artifacts-example/actions/workflows/solidity-review-artifacts.yml


scheibinger:
@Tofel Could fix the failed check `validate-worfklow-changes`

libs/jira-tracing/README.md
scheibinger, line 1: @Tofel Have you considered packaging this as a Typescript based action instead of a lib?

scheibinger (commented):
Looks good in general, just posted one question.

It's pretty complex setup, would be nice if you can add a CODEOWNER entry for all the files added in this PR, so we can easily tag the right people to review changes in it.

libs/jira-tracing/README.md
Tofel: I have not, because I have copied that code from `chainlink` repo. It was originally created by @momentmaker and @HenryNguyen5 and I didn't want to modify more things than strictly necessary.

scheibinger (approved):

PR #625, opened 2024-09-20
Update code owners for CTF actions

lukaszcl:


CODEOWNERS
```diff
@@ -1 +1,4 @@
-* @smartcontractkit/releng
\ No newline at end of file
+* @smartcontractkit/releng
+
+# Chainlink Testing Framework (CTF)
+/actions/ctf-* @smartcontractkit/test-tooling-team
```
lukaszcl, line 4: @scheibinger do you think this will work for all ctf- folders in /actions?

erikburt: I don't think so based on [these docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners).

```
# The `docs/*` pattern will match files like
# `docs/getting-started.md` but not further nested files like
# `docs/build-app/troubleshooting.md`.
docs/*  docs@example.com
```

scheibinger: From what I recall, GH uses GLOB patterns, I checked with https://globster.xyz/
And based on that you pattern wouldn't work.

Maybe you could try 
```suggestion
/actions/ctf-*/** @smartcontractkit/test-tooling-team
```

scheibinger (changes requested):
see the inline comment

CODEOWNERS
lukaszcl: thanks @scheibinger. I just applied your suggestion. Could you review again?

scheibinger (approved):

PR #660, opened 2024-10-08
Reduce noise from comments

chainchad:
See example [PR](https://github.com/smartcontractkit/chainlink/pull/14509) with some of the noise that this aims to address. This action should support editing an existing comment but for some reason it wasn't working all of the time. This refactors it to:

- Add PR comment ID to make lookups safer (instead of finding last comment from "github bot").
- Add workflow run ID in comment to make finding source of comments easier.
- Delete the previous PR comment and add a new one instead of editing last comment (this seemed to be buggy for force pushes (see above PR)).
- Optionally avoid PR commenting for success reports.

Future work to include:

- DRY this up and move inline shell code into separate file(s).

PR #665, opened 2024-10-10
action error reporter 0.3.0: Support multi-section + long log trimming + UX improvements

kalanyuz:
1. Fix issues where it writes over existing bot comments that are not from this script.
2. Added features that would improve user experience, handle larger log files
3. keep GH comment thread as clean as possible (ie. not posting success cases if no prior errors were found)


chainchad:
Another approach was in https://github.com/smartcontractkit/.github/pull/660 -- maybe we can consolidate or close one?

kalanyuz:
@chainchad of course, happy for you edit my PR too! Love that you add the link to the llm run itself too.

The script is designed to support subscribing to multiple workflows as well, hence the spamming behavior at first.
This edit fix should still allow multiple workflow while limiting its workspace to one comment.

For skipping success, I was thinking that leaving error as is even the CI/CD ran successfully might caused confusion, so ended up with just skipping only if the PR has no error runs.

See some test cases I did here ( lots of force pushes too): [success](https://github.com/smartcontractkit/aer_tester/pull/3), [single](https://github.com/smartcontractkit/aer_tester/pull/2), [multi](https://github.com/smartcontractkit/aer_tester/pull/4)

actions/llm-action-error-reporter/action.yml
```diff
@@ -142,56 +192,134 @@ runs:
...
         repo_url=$(gh repo view --json url --jq '.url')
+        wf_job_urls=$(gh run view ${{ github.event.workflow_run.id }} --json jobs --jq '.jobs | map("[\(.name)](\(.url))") | join(" , ")')
+        workflow_id=$(echo "$workflow_name" | tr -d ' ' | tr '[:upper:]' '[:lower:]')
 
-        pr_message="**Below is an analysis created by an LLM. Be mindful of hallucinations and verify accuracy.**
+        pr_message="
+        <$workflow_id>
 
-        ## WF: $workflow_name[#$short_sha]($repo_url/commit/$head_sha)
+        ## AER Report: [$workflow_name]($repo_url/actions/runs/${{ github.event.workflow_run.id }})
```
chainchad, line 205: Want to link to the specific job here? This link goes the run that called this workflow (the originating workflow).

Wondering if we should change this to use: `https://github.com/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}#${GITHUB_JOB})`

This should go to the exact run and job than ran this this `llm-action-error-report` action.

kalanyuz: @chainchad 
Thank you! Made some changes and resolved some DRY issues.

So this is how I imagined the script would work:
- if someone sees the error report then they may want to confirm the source
  - this will allow them to do so without scroll to the bottom or go to the actions tab
- links for specific jobs that errored out are provided as well to reduce a few more clicks
- now, the link to the AER workflow itself is probably not useful here
  - it wouldn't give out any information, just that OpenAI request is successful
  - and if the AER workflow fails then the comment wouldn't happen
  - but I've also added the link under the title just in case
  
Also noticed that `${GITHUB_RUN_ID}#${GITHUB_JOB}` url wouldn't work, as `$GITHUB_JOB` just returns the job name, and the URL gets redirected to the workflow run page directly. Directly linking to specific jobs seems to need `[run_id]/jobs/[job_number]`

Updated the example for the [test pr as well](https://github.com/smartcontractkit/aer_tester/pull/4).

actions/llm-action-error-reporter/action.yml
```diff
@@ -53,39 +82,49 @@ runs:
...
     - name: Checkout action repo
       if: ${{ env.SKIP_ACTION == 'false' }}
       uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
       with:
         repository: smartcontractkit/.github
         ref: ${{ inputs.workflow-ref }}
-        path: "error_reporter_actions"
-        # prevents access to any other action and workflow files
+        path:
+          "error_reporter_actions"
```
chainchad, line 95: Some weird formatting here where it's putting these values on a newline.

actions/llm-action-error-reporter/action.yml
```diff
@@ -53,39 +82,49 @@ runs:
...
+      env:
+        GH_TOKEN: ${{ inputs.gh-token }}
+        LINE_LIMIT: ${{ inputs.log-lines-limit }}
+      run: |
+        echo "Analyzing run ${{ github.event.workflow_run.id }}"
+        LOG_RAW=$(gh run view ${{ github.event.workflow_run.id }} --log-failed)
+        echo "$LOG_RAW" > rawlog.log
+
+        # trim logs and outputs to runlog.log
+        chmod +x ./error_reporter_actions/actions/llm-action-error-reporter/trim_logs.sh
```
chainchad, line 118: We can mark this as executable on the file system and git will account for it (if it's not already) to avoid having to set it here.

actions/llm-action-error-reporter/action.yml
```diff
@@ -53,39 +82,49 @@ runs:
...
+        ./error_reporter_actions/actions/llm-action-error-reporter/trim_logs.sh
+
     - name: Call ChatGPT
       if:
         ${{ env.SKIP_ACTION == 'false' && inputs.parent-workflow-conclusion ==
         'failure' }}
       shell: bash
       run: |
-        log_output=$(cat runlog.log | sed -e 's/\\ No newline at end of file//g' | sed -E 's/\\+$//g' | sed -E 's/\\+ //g')
+        log_output=$(cat runlog.log | jq -Rs . | sed 's/^"//;s/"$//' | sed -e 's/\\ No newline at end of file//g' | sed -E 's/\\+$//g' | sed -E 's/\\+ //g')
```
chainchad, line 127: It's not immediately obvious to me what these sed's are doing. Would be great to add a comment here to explain.

actions/llm-action-error-reporter/action.yml
```diff
@@ -123,28 +162,28 @@ runs:
...
+        </$WORKFLOW_ID>"
 
-        # Check if edit-comment is true and there are prior comments from github.actor
-        if [ "$edit_comment" = "true" ] && [ "$author_comments" -gt 0 ]; then
-          gh pr comment $PR_NUMBER -b "$pr_message" --edit-last
-        else
-          gh pr comment $PR_NUMBER -b "$pr_message"
-        fi
+        # post the error report to the PR
+        chmod +x ./error_reporter_actions/actions/llm-action-error-reporter/update_pr_comment.sh
```
chainchad, line 185: See other comment about making this executable

actions/llm-action-error-reporter/action.yml
```diff
@@ -154,23 +193,36 @@ runs:
...
-        # Check if edit-comment is true and there are prior comments from github.actor
-        if [ "$edit_comment" = "true" ] && [ "$author_comments" -gt 0 ]; then
-          gh pr comment $PR_NUMBER -b "$pr_message" --edit-last
-        else
-          gh pr comment $PR_NUMBER -b "$pr_message"
-        fi
+        </$WORKFLOW_ID>"
+
+        # post the error report to the PR
+        chmod +x ./error_reporter_actions/actions/llm-action-error-reporter/update_pr_comment.sh
```
chainchad, line 215: See other comments

actions/llm-action-error-reporter/action.yml
```diff
@@ -154,23 +193,36 @@ runs:
...
+        ./error_reporter_actions/actions/llm-action-error-reporter/update_pr_comment.sh
+
+    - name: Collect metrics
```
chainchad, line 218: We're getting rid of this action so we can remove `Collect metrics` completely.

actions/llm-action-error-reporter/aer.sh
```diff
@@ -4,24 +4,31 @@
 input=$(cat)
```
chainchad, line 4: I'd set the shebang on all of these scripts like this along with setting it to fail on errors, unassigned vars, and when a piped command fails like:

```
#!/usr/bin/env bash

set -euo pipefail
```

actions/llm-action-error-reporter/trim_logs.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
chainchad, line 1:
```suggestion
#!/usr/bin/env bash
```

actions/llm-action-error-reporter/trim_logs.sh
```diff
@@ -0,0 +1,50 @@
...
+set -euo pipefail
```
chainchad, line 2: 👍

actions/llm-action-error-reporter/update_pr_comment.sh
```diff
@@ -0,0 +1,59 @@
+
+#!/bin/bash
```
chainchad, line 2:
```suggestion
#!/usr/bin/env bash
```

actions/llm-action-error-reporter/action.yml
kalanyuz: Did not know about this before. Very neat!

actions/llm-action-error-reporter/aer.sh
chainchad, line 1: Where do we call this script?

actions/llm-action-error-reporter/action.yml
```diff
@@ -154,23 +173,23 @@ runs:
...
-        # Check if edit-comment is true and there are prior comments from github.actor
-        if [ "$edit_comment" = "true" ] && [ "$author_comments" -gt 0 ]; then
-          gh pr comment $PR_NUMBER -b "$pr_message" --edit-last
-        else
-          gh pr comment $PR_NUMBER -b "$pr_message"
-        fi
+        </$WORKFLOW_ID>"
+
+        # post the error report to the PR
+        ./error_reporter_actions/actions/llm-action-error-reporter/update_pr_comment.sh
```
chainchad, line 195: Is this path correct? Here's an example of how we can refer to the action path dynamically: 

https://github.com/smartcontractkit/.github/blob/a40ee26a986e00fbef228ded7e7da9bfed64dd2c/actions/cicd-build-publish-charts/action.yml#L93

actions/llm-action-error-reporter/action.yml
```diff
@@ -123,28 +143,27 @@ runs:
...
+        </$WORKFLOW_ID>"
 
-        # Check if edit-comment is true and there are prior comments from github.actor
-        if [ "$edit_comment" = "true" ] && [ "$author_comments" -gt 0 ]; then
-          gh pr comment $PR_NUMBER -b "$pr_message" --edit-last
-        else
-          gh pr comment $PR_NUMBER -b "$pr_message"
-        fi
+        # post the error report to the PR
+        ./error_reporter_actions/actions/llm-action-error-reporter/update_pr_comment.sh
```
chainchad, line 166: Is this path correct? Here's an example of how we can refer to the action path dynamically: 

https://github.com/smartcontractkit/.github/blob/a40ee26a986e00fbef228ded7e7da9bfed64dd2c/actions/cicd-build-publish-charts/action.yml#L93

actions/llm-action-error-reporter/action.yml
```diff
@@ -76,16 +76,36 @@ runs:
...
+      env:
+        GH_TOKEN: ${{ inputs.gh-token }}
+        LINE_LIMIT: ${{ inputs.log-lines-limit }}
+      run: |
+        echo "Analyzing run ${{ github.event.workflow_run.id }}"
+        LOG_RAW=$(gh run view ${{ github.event.workflow_run.id }} --log-failed)
+        echo "$LOG_RAW" > rawlog.log
+
+        # trim logs and outputs to runlog.log
+        ./error_reporter_actions/actions/llm-action-error-reporter/trim_logs.sh
```
chainchad, line 98: Is this path correct? Here's an example of how we can refer to the action path dynamically: 

https://github.com/smartcontractkit/.github/blob/a40ee26a986e00fbef228ded7e7da9bfed64dd2c/actions/cicd-build-publish-charts/action.yml#L93

actions/llm-action-error-reporter/aer.sh
kalanyuz: Not in this github action. It's offered as a compliment tool that you could try things out in CLI: https://github.com/smartcontractkit/.github/pull/665#pullrequestreview-2366644348

actions/llm-action-error-reporter/action.yml
kalanyuz: @chainchad
The path is correct and uses the sparse-checkout path ( tested with rev in the test repository .)
Changed to go with your suggestion as these scripts aren't meant to be customizable like the prompt file.

chainchad (approved):

PR #674, opened 2024-10-16
Add ability to set multiple report paths in SonarQube actions

chudilka1:


actions/ci-sonarqube-ts/action.yml
```diff
@@ -80,32 +80,32 @@ runs:
...
-            inputs.typescript-lcov-report-path
-          }}
-      env:
-        SONAR_TOKEN: ${{ inputs.sonar-token }}
-        SONAR_HOST_URL: ${{ inputs.sonar-host-url }}
+    - name: Set SonarQube Report Paths
+      id: sonarqube_report_paths
+      shell: bash
+      run: |
+        sonarqube_coverage_report_paths=$(find -type f -name '${{ inputs.typescript-lcov-report-path }}' -printf "%p,")
```
scheibinger, line 87: What does this line do?
What is expected fromat for passing multiple paths in the input? Is it a string with newlines as delimiter?

scheibinger: it would be nice if you can document what is expected format of the input parameter in the input description

scheibinger (commented):
I left an inline comment

actions/ci-sonarqube-ts/action.yml
chudilka1: It recursively looks for the file names specified in action input `typescript-lcov-report-path` (that corresponds to the report name) and assigns those values to this variable, used later by Sonar. It is tested [here](https://github.com/smartcontractkit/capabilities/actions/runs/11381438006/job/31662910820?pr=33#step:3:157), you may find paths delimited by coma (`%p,` where `p` is a path/value found)

chudilka1: I have updated the input variable names, so users may now specify only report names instead of paths (the latter will be found automatically). If users specify paths, the action will fail, as it will look for file names.

scheibinger: Thanks for explaining, so it assumes that all packages, modules in the project produce a report with the same file name? right?

It probably helps for some use cases, but for the other use cases the user of the workflow is not able to specify explicitly which exact report paths should be used. Can this be a problem for mono repos with variety of different frameworks and tools for generating such reports?

scheibinger (approved):
Thanks for the answers on my questions, now it is all clear! LGTM

chudilka1 (commented):
@scheibinger
I have added [here](https://github.com/smartcontractkit/.github/pull/674/files#diff-cd2ed450d7e06fa8063ee556f4d85055d23e6c461f142773801b2767f7c264c3R61) the validation of conditions: when one of the workflow names is not specified, trigger the other action that downloads artifacts from the same workflow. When the workflow name that generates a report is not specified, it always means that the action is used by the same workflow that produces reports. How it works may be found in two different repos, that use different inputs:
1. [chainlink-common](https://github.com/smartcontractkit/chainlink-common/pull/849/files) - different workflows generate reports and are fetched from different workflows
2. [capabilities](https://github.com/smartcontractkit/capabilities/pull/33) (monorepo) - fetches different reports from different places, produced by the same workflow, where this action is specified.

scheibinger (approved):

PR #678, opened 2024-10-22
checks for changesetKey if passed in

momentmaker:
This checks for a changesetKey in the changesetFile if `process.env.CHANGESET_KEY` is passed in.

This should enforce that the keys are correct per different changeset directory.

### DIFF
--- a/.changeset/brown-trainers-type.md
+++ b/.changeset/brown-trainers-type.md
@@ -1,0 +1,5 @@
+---
+"jira-tracing": minor
+---
+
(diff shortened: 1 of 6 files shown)


erikburt (commented):
Should some tests be added for this functionality?

momentmaker:
ahh i had tested here:
https://github.com/smartcontractkit/chainlink/pull/14894

i didn't see a test file for the changeset-libs

erikburt (approved):

PR #709, opened 2024-11-16
action: go conditional tests

erikburt:
Adds a nodejs action designed to conditionally run unit tests. It does this by:

1. Building all test binaries
2. Hashing all test binaries
3. Comparing the hash of the recently built test binaries, to that of a source-of-truth
4. Run only the test binaries who's hashes have changed

See [README](https://github.com/smartcontractkit/.github/blob/feat/go-test-binary-comparison/apps/go-test-caching/README.md) for more information.

---

RE-3168


apps/go-test-caching/src/main.ts
```diff
@@ -0,0 +1,191 @@
...
+    buildFlags,
+    hashesBranch,
+    hashesFile: `${testSuite}.json`,
+    testSuite,
+    runAllTests,
+    collectCoverage,
+    forceUpdateIndex,
+  };
+}
+export async function run() {
```
erikburt, line 81: The entrypoint to the action is here.

apps/go-test-caching/README.md
```diff
@@ -0,0 +1,177 @@
+# go-test-caching
```
erikburt, line 1: Also probably going to rename this. Any recommendations?

apps/go-test-caching/src/pipeline.ts
```diff
@@ -0,0 +1,304 @@
...
+/**
+ * STEP 2: Build Test Binaries
+ */
+export async function buildTestBinaries(
+  inputs: Inputs,
+  packages: LocalPackages,
+): Promise<CompiledPackages> {
+  logSection("Build Tests");
+
+  const maxBuildConcurrency = parseInt(core.getInput("build-concurrency")) || 4;
```
jmank88, line 82: The yaml default is `8` but here we have `4`. Should they be aligned? Could they be dynamic like GOMAXPROCS rather than hard-coded?

apps/go-test-caching/README.md
sebawo: `go-run-conditional-tests` or `go-conditional-tests` ?

apps/go-test-caching/src/pipeline.ts
erikburt: Defaulted both run/build concurrency to number of CPU cores.

erikburt: 5c582558ed60cee185a7f1b8c70c29b829263a72

apps/go-test-caching/README.md
erikburt: renamed to `go-conditional-tests`

apps/go-conditional-tests/README.md
```diff
@@ -0,0 +1,177 @@
...
+### General Inputs
+
+- `test-suite`
+  - The name of the test suite, used to scope artifacts and the test indexes
+- `module-directory`, path (`./`)
+  - The path to the root module for the tests. Similar to setting
+    `working-directory`.
+- `hashes-branch`, string (`test-hashes`)
+  - The (ideally orphaned) git branch to store the test hash index json files
+    on. Used by `run` and `update`.
```
erikburt, line 136: Add comment that this needs to be setup prior to running.

chainchad (approved):

PR #711, opened 2024-11-20
Improving Envoy config and add support for debugging

njegosrailic:
## What 

- Use `LOGICAL_DNS` instead of `STRINCT_DNS` - I think this makes more sense. See: https://abhishekvrshny.medium.com/dns-aware-persistent-connections-abda9921db34
- Configure  keepalive, idle_timeout and  max_connection_duration. I'm not confident that these values are good long term, but we can tune them later. I have tested them and no issues.

Config parameters changed in this PR:

- `idle_timeout`: Specifies the duration of inactivity after which a connection will be closed to conserve resources.
- `max_connection_duration`: Defines the maximum allowable lifetime of a connection, regardless of activity, to ensure stability.
- `connection_keepalive.interval`: Sets the frequency at which keepalive pings are sent to maintain the connection.
- `connection_keepalive.timeout`: Determines how long to wait for a response to a keepalive ping before marking the connection as failed.

## Why 

- Fix the issues with connection reset.

### DIFF
--- a/.changeset/selfish-wombats-fetch.md
+++ b/.changeset/selfish-wombats-fetch.md
@@ -1,0 +1,5 @@
+---
+"setup-gap": minor
+---
+
+Improving Envoy config and add support for debugging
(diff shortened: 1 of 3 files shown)


chainchad (commented):
Quick suggestion but looks great otherwise

chainchad (approved):

PR #717, opened 2024-11-22
Adding ECR login to crib-deploy-environment action

njegosrailic:
## What 

- Adding ECR login to the crib-deploy-environment action, required for Helm. Deploying CRIB environments is not possible without this step. Previously, ECR login was only required for the aws-sigv4-proxy image, but now that it’s no longer used, an explicit login step is needed. 
- Bumping the setup-gap action version. The new version includes a fix for connection timeouts and supports passing extra inputs.
- Removed the `KUBECACHEDIR` environment variable, as it is not needed for GAP v2.
- Adding collect Envoy proxy logs step. We have the same step in `setup-gap` actions but it will collect only initial logs so we need it here as well. 

## Why

- Improvements and fixes in order to make CRIB using GAP v2 more reliable.

### DIFF
--- a/.changeset/small-chairs-care.md
+++ b/.changeset/small-chairs-care.md
@@ -1,0 +1,5 @@
+---
+"crib-deploy-environment": minor
+---
+
+Adding ECR login required for Helm.
--- a/.changeset/small-chairs-care.md
+++ b/.changeset/small-chairs-care.md
@@ -1,5 +1,5 @@
 ---
-"crib-deploy-environment": minor
+"crib-deploy-environment": major
 ---
 
 Adding ECR login required for Helm.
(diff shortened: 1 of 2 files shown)


.changeset/small-chairs-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"crib-deploy-environment": minor
```
rafaelfelix, line 2: I'd argue this should be a major version jump, specially because of the new required input `aws-ecr-private-registry`

rafaelfelix (changes requested):
looks good, just a comment on the versioning of it.

.changeset/small-chairs-care.md
njegosrailic: Thanks. Fixed.

rafaelfelix (approved):

PR #719, opened 2024-11-22
CCIP-4303: Setting postgres instance for integration tests

b-gopalswami:
In order to have in-memory test in integration-tests workflow. We need PG instances. This PR is setting up the separate workflow for integration-tests with postgres service containers enabled.


.github/workflows/run-e2e-tests.yml
```diff
@@ -590,12 +590,21 @@ jobs:
...
         inputs.test_config_override_path }}
       TEST_ID: ${{ matrix.tests.id_sanitized || matrix.tests.id }}
+      DB_URL: postgresql://postgres:postgres@localhost:5432/chainlink_test?sslmode=disable
     steps:
       - name: Collect Test Telemetry
         if: inputs.collect_test_telemetry
         uses: catchpoint/workflow-telemetry-action@94c3c3d9567a0205de6da68a76c428ce4e769af1 # v2.0.0
       - name: Checkout repository
         uses: actions/checkout@v4.2.1
+      - name: Setup postgres
```
kalverra, line 600: I think we should separate this out into its own workflow instead of bolting it onto the E2E tests.

kalverra (changes requested):

.github/workflows/run-integration-tests.yml
```diff
@@ -0,0 +1,676 @@
...
+        required: false
+      # Used in some tests to send slack notifications
+      SLACK_API_KEY:
+        required: false
+      # Used in some tests to send slack notifications
+      SLACK_CHANNEL:
+        required: false
+
+env:
+  CHAINLINK_IMAGE:
```
kalverra, line 171: If these are more traditional integration tests, are we still using Chainlink docker images? I assumed these would be in-memory?

actions/setup-postgres/docker-compose.yml
kalverra, line 1: Check out [GHA service containers](https://docs.github.com/en/actions/use-cases-and-examples/using-containerized-services/about-service-containers), which should handle this sort of thing with less overhead in GHA.

kalverra (changes requested):

kalverra (dismissed):

.github/workflows/run-integration-tests.yml
```diff
@@ -0,0 +1,404 @@
...
+            echo "No dynamic environment variables for $test_id."
+          else
+            echo "$json_content" | jq -r 'to_entries | .[] | "\(.key)=\(.value)"' | while IFS='=' read -r key value; do
+              echo "Setting $key=$value for $test_id"
+              echo "$key=$value" >> "$GITHUB_ENV"
+            done
+          fi
+      - name: Run tests
+        id: run_tests
+        uses: smartcontractkit/.github/actions/ctf-run-tests@b8731364b119e88983e94b0c4da87fc27ddb41b8 # ctf-run-tests@0.0.0
```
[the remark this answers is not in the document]
b-gopalswami: As discussed offline, will improvise this with follow up PR.

chainchad (approved):

PR #721, opened 2024-11-24
Adding support for using a custom header for passing Github OIDC JWT Auth token

njegosrailic:
## What 

See title. 

## Why 

We would like to use a custom header to avoid interfering with the default Authorization header, which would require changes on the app side. Also, sorting out the list of the ENVs.


rafaelfelix (approved):
🚀

PR #725, opened 2024-11-25
setting up go before running CRIB (ensures GOBIN is in place)

rafaelfelix:
this should fix the issue related to the crib executable not being found in PATH, as well as caching dependencies

closes [CRIB-561 ](https://smartcontract-it.atlassian.net/browse/CRIB-561)


rafaelfelix:
reopening as we'll need this fix in later versions of the action

chainchad (approved):

PR #727, opened 2024-11-26
Adding support for dynamically fetching the JWT token for Kubernetes API access 

njegosrailic:
## What 

See title. 

## Why 

In some cases, requests may run for a long time, and the JWT might need to be refreshed. In the current Envoy configuration, we fetch the token before the proxy starts, and there is no way to refresh it afterward.

https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter#httpcall
> asynchronous is a boolean flag. If async is set to true, Envoy will make the HTTP request and continue, regardless of the response success or failure. If this is set to false, or not set, Envoy will suspend executing the script until the call completes or has an error.

Note: I added `json.lua` file in order to be able to parse JSON properly since the lib isn't available and in order to avoid creating a custom Envoy image. Also, calls from LUA to external services are only possible over envoy cluster.

### DIFF
--- a/actions/setup-gap/envoy.yaml.template
+++ b/actions/setup-gap/envoy.yaml.template
@@ -26,11 +26,71 @@ static_resources:
                     typed_config:
                       "@type": type.googleapis.com/envoy.extensions.filters.http.lua.v3.Lua
                       inline_code: |
-                        function envoy_on_request(request_handle)
-                          request_handle:headers():remove("host")
-                          request_handle:headers():add("host", "${K8S_API_ENDPOINT}")
-                          request_handle:headers():add("${ENVOY_GITHUB_OIDC_TOKEN_HEDER_NAME}", "Bearer ${JWT_TOKEN}")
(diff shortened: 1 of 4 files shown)


actions/setup-gap/envoy.yaml.template
```diff
@@ -26,11 +26,71 @@ static_resources:
...
+                            -- Set the audience and the GitHub OIDC token URL
+                            local audience = "gap"
+                            local oidc_url = "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=" .. audience
+
+                            -- Prepare headers
+                            local headers = {
+                                [":method"] = "GET",
+                                [":path"] = oidc_url,
+                                [":authority"] = "${GITHUB_OIDC_HOSTNAME}",
+                                ["Authorization"] = "Bearer " .. "${ACTIONS_ID_TOKEN_REQUEST_TOKEN}", -- Use environment variable
```
chainchad, line 51: Just wanted to confirm that this env var is available within the running container?

PR #737, opened 2024-12-03
Use go version from cli/go.mod

scheibinger:
This is required for https://github.com/smartcontractkit/crib/pull/253 to work
Tested here: https://github.com/smartcontractkit/crib/actions/runs/12141161594

### DIFF
--- a/actions/crib-deploy-environment/action.yml
+++ b/actions/crib-deploy-environment/action.yml
@@ -141,7 +141,7 @@ runs:
 
     - uses: actions/setup-go@v5
       with:
-        go-version-file: "${{ github.workspace }}/crib/go.work"
+        go-version-file: "${{ github.workspace }}/crib/cli/go.mod"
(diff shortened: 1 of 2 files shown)


actions/crib-deploy-environment/action.yml
```diff
@@ -141,7 +141,7 @@ runs:
 
     - uses: actions/setup-go@v5
       with:
-        go-version-file: "${{ github.workspace }}/crib/go.work"
+        go-version-file: "${{ github.workspace }}/crib/cli/go.mod"
```
rafaelfelix, line 144: should we change `cache-dependency-path` as well? just to keep both in sync

rafaelfelix: nevermind, it actually makes sense to keep a cache of all golang apps inside the repo 👍

rafaelfelix (approved):

PR #760, opened 2024-12-16
Pass GAP_GH_JWT_TOKEN env var

scheibinger:


actions/crib-deploy-environment/action.yml
```diff
@@ -236,6 +237,7 @@ runs:
         PROFILES: ${{ inputs.devspace-profiles }}
         SETUP_AWS_PROFILE: false
         SETUP_EKS_CONFIG: false
+        GAP_GH_JWT_TOKEN: ${{ steps.setup-gap-k8s.outputs.gh-jwt-token }}
```
njegosrailic, line 240: Hey, sorry for the confusion, but for the K8s API, we fetch the token dynamically. If you would like to access other services, you would need to implement that on the app/CLI side.

I have added a PR to clean up the ENV variable: https://github.com/smartcontractkit/.github/pull/761

PR #775, opened 2024-12-30
Add a local dynamic proxy server config

njegosrailic:
## What

- Added a local dynamic proxy server configuration.
- Renamed some input parameters to avoid confusion.
- Added extra input parameters.
- Added validation for all required environment variables (ENVs) in the Envoy template, as `envsub` does not fail. Clear and concise error messages will be displayed if something is missing.
- Modified the "Verify Envoy Proxy" step to support multiple listeners.
- Modified the LUA script to allow specifying the JWT header name instead of using a hardcoded name. This was a bug.

## Why

The primary goal is to set up a local HTTPS proxy to proxy all requests and dynamically append the GitHub OIDC JWT token to a specific header. This will simplify the process, as most languages and software respect the `HTTPS_PROXY` ENV variable, and no need to modify the app code to fetch and append the header. 

However, we may encounter other issues, such as SSL certificate validation. The local CA used to issue a self-signed TLS certificate should already be added to the main CA chain, but there may be other challenges. This is just the initial version, so please bear with me.

Once we configure this ENV, all requests will be sent through the dynamic forward proxy, so we need to be cautious, as we will proxy all app requests. We must ensure that the JWT token is not added to every request (for example other public domains). This is why I've added the `main-dns-zone` input. We can discuss this internally - perhaps we should just add a zone, and developers can use prefixes for service names. (We already have the DNS zone name in the K8s API hostname, so what I'm proposing is having just one secret for the DNS zone and using prefixes.)

During debugging, I noticed that missing ENVs can cause Envoy to crash, making it difficult to identify issues. By explicitly validating all required ENVs, it could save time in the future.

One more thing: I know that the LUA script is almost identical for both listeners, resulting in duplication, and it could be moved to a separate file and loaded. However, there are some loops involved in request processing, and I'd prefer to avoid altering that for now. We can improve this later, especially since I don't expect us to add more listeners anytime soon.


actions/setup-gap/action.yml
```diff
@@ -190,23 +208,32 @@ runs:
       if: inputs.use-k8s == 'true'
       shell: sh
       env:
+        DYNAMIC_PROXY_PORT: ${{ inputs.dynamic-proxy-port }}
         ENABLE_PROXY_DEBUG: ${{ inputs.enable-proxy-debug }}
-        ENVOY_LOG_LEVEL: "info"
+        GITHUB_OIDC_TOKEN_HEDER_NAME:
```
chainchad, line 213:
```suggestion
        GITHUB_OIDC_TOKEN_HEADER_NAME:
```

Would also need to update all refs to this new name.

actions/setup-gap/envoy.yaml.template
```diff
@@ -127,6 +126,150 @@ static_resources:
...
+                                nil, -- No body for GET requests
+                                3000 -- Timeout in milliseconds
+                            )
+
+                            if response_headers[":status"] == "200" then
+                                request_handle:logInfo("GitHub OIDC token fetched successfully.")
+                                request_handle:logInfo(response_body)
+                                local data = JSON.decode(response_body)
+                                jwt_token = data.value -- Assign the token value
+                                jwt_expiration = os.time() + 300 -- Tokens are valid for 5 minutes
```
chainchad, line 223: Would it make sense to get this dynamically from the JWT claims?

chainchad (commented):
Looks great, just a spelling thing and a question.

actions/setup-gap/envoy.yaml.template
njegosrailic: Makes sense. Let me try that. 👍

actions/setup-gap/envoy.yaml.template
```diff
@@ -61,31 +110,32 @@ static_resources:
...
                             end
 
                             refreshing = false
                         end
 
                         -- Function to refresh the token if needed
                         function refresh_token_if_needed(request_handle)
-                            local current_time = os.time()
-                            if not jwt_token or jwt_expiration - current_time < 60 then
+                            if not jwt_token or jwt_expiration - os.time() < 60 then
```
njegosrailic, line 126: @chainchad I'm unsure about refreshing the token if it is about to expire in less than 60 seconds. Maybe it's too frequent? Thoughts?

chainchad (approved):

PR #786, opened 2025-01-04
Clearing Python-2.0

afelisatti:


chainchad (approved):

chainchad (changes requested):
Just needs a .changeset/ markdown file (run `pnpm changeset` to bump the version of this (patch is fine)) and it should be good 👍

PR #804, opened 2025-01-14
Migrating E2E Tests workflow to GAP v2

njegosrailic:
## What 

See title. 

## Why 

We are migrating to the latest version of `ctf-run-tests` GHA that uses GAP for accessing K8s API.


rafaelfelix (commented):
~~correct me if I'm wrong, but from the changes made to inputs we probably need a major version bump~~
UPDATE: versioning applies for actions, not workflows

rafaelfelix (approved):

lukaszcl (approved):

PR #823, opened 2025-01-20
Add all subdomains of the selected dns zone as cert SANs

JooKop:
By adding all subdomains of the selected dns zone as cert SANs, clients that can't be configured with custom host headers or insecure connections can still utilize the proxy.

If you set, e.g. `127.0.0.1 my-service.my-dns-zone` in the GHA runner's `/etc/hosts` file, any client can then directly connect to e.g. `https://my-service.my-dns-zone` with full functionality.

### DIFF
--- a/.changeset/wildcard-sans.md
+++ b/.changeset/wildcard-sans.md
@@ -1,0 +1,11 @@
+---
+"setup-gap": patch
+---
+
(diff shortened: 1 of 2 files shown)


DimitriosNaikopoulos (approved):
LGTM ✅ I don't have a lot of experience with GAPv2 so you may want to get another review or merge and revert if there is anything that I did not catch but the approach and the [testing](https://github.com/smartcontractkit/releng-test/actions/runs/12870646373/workflow?pr=56) looks good

PR #824, opened 2025-01-20
[Infoplat-1563] Validate protobuf schema evolution compatibility using schema registry

anukin:
## What
This adds a new action which integrates beholder validator for schema changes

## Why?
This is needed so that beholder validator can be integrated with the schema changes given a beholder.yaml is present in the repo.


actions/ci-beholder-validator/README.md
```diff
@@ -0,0 +1,164 @@
+# Schema Evolution Validator
```
anukin, line 1: This is not accurate and will be changed once the testing can be completed satisfactorily

actions/ci-beholder-validator/README.md
```diff
@@ -0,0 +1,164 @@
...
+  check-schemas:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
```
erikburt, line 69: Did you mean v4 specifically?

```suggestion
        uses: actions/setup-go@v5
```

actions/ci-beholder-validator/README.md
```diff
@@ -0,0 +1,164 @@
...
+      schema: "./schemas/on_ramp_send.proto"
+    - entity: OffRampSend
+      schema: "./schemas/off_ramp_send.avsc"
+```
+
+Note: Schema paths should be relative to the `beholder.yaml` location.
+
+### A sample setup without using actions folder
+
+1. Create `.github/workflows/schema-check.yml`:
```
erikburt, line 47: Would it make sense to make a [reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows), which then calls this action? It would allow you to easier update the consumers of this action/workflow.

Both options work well though so no pressure.

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,101 @@
...
+
+inputs:
+  checkout-repo:
+    description: "Whether to checkout the repository"
+    required: false
+    default: "true"
+  checkout-repo-fetch-depth:
+    description: "Fetch depth for checkout"
+    required: false
+    default: "0"
```
erikburt, line 12: You can remove this and the first step of the action if you want. It doesn't really add much value, especially if you're checking out the repo before calling this action anyways.

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,101 @@
...
+      if: inputs.checkout-repo == 'true'
+      uses: actions/checkout@v4.2.1
+      with:
+        fetch-depth: ${{ inputs.checkout-repo-fetch-depth }}
+    - name: Detect changed files
+      id: changes
+      shell: bash
+      run:
+        echo "CHANGED_FILES=$(git diff --name-only origin/master)" >>
+        $GITHUB_ENV
```
erikburt, line 45: We typically use this action for changed files: https://github.com/dorny/paths-filter/

Here's a large example of usage: https://github.com/smartcontractkit/chainlink/blob/develop/.github/workflows/ci-core.yml#L20-L110

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,101 @@
...
+      with:
+        role-to-assume: ${{ inputs.aws-role-arn }}
+        role-duration-seconds: ${{ inputs.aws-role-duration-seconds }}
+        aws-region: ${{ inputs.aws-region }}
+        mask-aws-account-id: true
+    - name: Login to aws ecr
+      if: inputs.docker-registry == 'aws'
+      uses: aws-actions/amazon-ecr-login@062b18b96a7aff071d4dc91bc00c4c1a7945b076 # v2.0.1
+      with:
+        registries: ${{ steps.process-params.outputs.aws-account-number }}
```
erikburt, line 62: What step is this referencing?

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,101 @@
...
+    - name: Pull schema-validator image from ECR
+      shell: bash
+      run: |
+        docker pull ${{ steps.login-ecr.outputs.registry }}/atlas-beholder-schema-validator:latest
```
erikburt, line 66: Could make the image tag an input that defaults to latest?

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,101 @@
...
+        --config-file /app/repo/config.yaml
+    - name: Validate schemas in master branch
+      if: steps.set_mode.outputs.mode == 'pr'
+      shell: bash
+      run: |
+        # Store current branch
+        CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+        # Checkout and validate master
+        git checkout master
```
erikburt, line 83: Can use the default branch? `${{ github.event.repository.default_branch }}`, or have it as an input.

(`origin/master` also referenced above)

actions/ci-beholder-validator/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "ci-beholder-validator",
+  "version": "0.0.0",
```
erikburt, line 3: Please add a changeset before merging this, it will bump this version and create a git tag for you.

```
pnpm install
pnpm changeset
```

erikburt (commented):
Looks good, only some suggestions.

actions/ci-beholder-validator/action.yml
anukin: I went through the documentation and this seems to be extremely useful when you want changes with some filter condition. 
In my case, we just need to provide changed files and the actual filtering is done by the code. This is the relevant piece of code https://github.com/smartcontractkit/atlas/blob/7acd79f562164d05f7b1d389ba74b40248a574d7/beholder/schema_validator/cmd/validate.go#L62

Does it make sense to change this in that case? LMK what you think?

actions/ci-beholder-validator/action.yml
anukin: will do

actions/ci-beholder-validator/action.yml
anukin: Will remove this. This was taken from an example(build-and-push) of this repo for seeing how ECR login worked. This is not needed in our case

actions/ci-beholder-validator/action.yml
anukin: I was not aware of that possibility(mainly tested this locally). Will make this change.

actions/ci-beholder-validator/package.json
anukin: will do.

actions/ci-beholder-validator/action.yml
erikburt: No that's fine, your solution is good.

actions/ci-beholder-validator/README.md
```diff
@@ -0,0 +1,115 @@
...
+      - "**.proto"
+      - "**.avsc"
+      - "**/beholder.yaml"
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: smartcontractkit/ci-beholder-validator@v1
```
erikburt, line 40: If this is supposed to reference this action it will look more like:

`uses: smartcontractkit/.github/actions/ci-beholder-validator@<sha> # ci-beholder-validator@1.0.0`

But if you're hoping to have a @v1 tag then you will need a separate repository that syncs this action

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,93 @@
...
+    description: ""
+    required: false
+
+runs:
+  using: composite
+  steps:
+    - name: Checkout repo
+      uses: actions/checkout@v4.2.1
+      with:
+        fetch-depth: ${{ inputs.checkout-repo-fetch-depth }}
```
erikburt, line 34: 1. Your README says to put a checkout before calling this action. If that's the case then this is unnecessary
2. This input doesn't exist

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,93 @@
...
+    - name: Pull schema-validator image from ECR
+      shell: bash
+      run: |
+        docker pull ${{ steps.login-ecr.outputs.registry }}/atlas-beholder-schema-validator:${{ inputs.image-tag }}
+    - name: Run Schema Validator
+      shell: bash
+      run: |
+        docker run -v "$PWD:/app/repo" \
+        --network redpanda_network \
+        -e SCHEMA_REGISTRY_URL=http://redpanda:8081 \
```
erikburt, line 65: Should the hostname be `redpanda-0` as per the docker-compose?

erikburt (changes requested):

actions/ci-beholder-validator/README.md
anukin: Let me change this

actions/ci-beholder-validator/action.yml
anukin: I changed this. So this is not needed. Thank you will do this

actions/ci-beholder-validator/README.md
anukin: Does the new change look okay? In some cases I am only seeing sha and in some cases I am seeing versions.

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,93 @@
...
+      id: changes
+      shell: bash
+      run:
+        echo "CHANGED_FILES=$(git diff --name-only origin/master)" >>
+        $GITHUB_ENV
+    - name: Start Redpanda with Docker Compose
+      shell: bash
+      run: |
+        docker compose up -d redpanda-console
+    - name: Configure aws creds
```
anukin, line 45: @erikburt I would love to have your review here. I am not that familiar with this. Will this approach to pulling images work in other repos where these actions would be used?
Are there some changes/alternatives I should explore?

actions/ci-beholder-validator/README.md
erikburt: The tag that will be created in this repo will be `ci-beholder-validator@1.0.0`. Which means the reference will need to be:

```
smartcontractkit/.github/actions/ci-beholder-validator@<sha> # ci-beholder-validator@1.0.0
```

It will need to use a sha because you cannot reference a tag with an `@`

actions/ci-beholder-validator/action.yml
erikburt: Yes this should work.

The callers of this action will need to pass an AWS role they can assume, and that role requires read-access to the ECR.

actions/ci-beholder-validator/README.md
anukin: Done

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,105 @@
...
+      run: |
+        docker pull ${{ steps.login-ecr.outputs.registry }}/atlas-beholder-schema-validator:${{ inputs.image-tag }}
+    - name: Validate schemas in master branch
+      shell: bash
+      run: |
+        # Store current branch
+        CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+        # Checkout and validate default branch
+        git checkout ${{ github.event.repository.default_branch }}
```
DimitriosNaikopoulos, line 87: Could this be set as an env variable eg 
```
shell: bash
env:
  DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
```

and then used like this
```suggestion
        git checkout ${{ env.DEFAULT_BRANCH }}
```
Issue [attacker-controllable code ](https://openssf.org/blog/2024/08/12/mitigating-attack-vectors-in-github-workflows/#:~:text=Code%20injection%20on%20GitHub%20Actions,permissions%20granted%20to%20the%20workflow.).

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,105 @@
...
+  # general inputs
+  checkout-repo-fetch-depth:
+    description: "number of commits to fetch"
+    required: false
+    default: "0"
+  # image inputs
+  docker-registry:
+    description: "Docker registry to use"
+    required: true
+    default: "aws"
```
DimitriosNaikopoulos, line 14: If this is required then does it need a default? Since the default it will always be overridden by the user's value 🤔

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,105 @@
...
+      uses: actions/checkout@v4.2.1
+      with:
+        fetch-depth: 0
+    - name: Detect changed files
+      id: changes
+      shell: bash
+      run: |
+        git fetch --prune origin +refs/heads/*:refs/remotes/origin/*
+        # TODO: This logic need to be moved to code
+        changed_files=$(git diff --name-only origin/${{github.event.repository.default_branch }})
```
DimitriosNaikopoulos, line 47: Could this be set as an env variable eg 
```
shell: bash
env:
  DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
```

and then used like this
```suggestion
        changed_files=$(git diff --name-only origin/${{ env.DEFAULT_BRANCH }})
```
Issue [attacker-controllable code ](https://openssf.org/blog/2024/08/12/mitigating-attack-vectors-in-github-workflows/#:~:text=Code%20injection%20on%20GitHub%20Actions,permissions%20granted%20to%20the%20workflow.).

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,105 @@
...
+        prefixed_files="${prefixed_files#,}"
+
+        echo "CHANGED_FILES=$prefixed_files" >> "$GITHUB_ENV"
+    - name: Start Redpanda with Docker Compose
+      shell: bash
+      run: |
+        docker compose -f "${{ github.action_path }}/docker-compose.yml" up -d redpanda-console
+    - name: Configure aws creds
+      uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
+      if: inputs.docker-registry == 'aws'
```
DimitriosNaikopoulos, line 64: Just thinking out loud here but is there a scenario that we will not need to pull the docker image? Because the following steps are also using the input from the `Login to aws ecr` step therefore we may not need to perform this check 🤔

DimitriosNaikopoulos (changes requested):

actions/ci-beholder-validator/action.yml
anukin: Hey @DimitriosNaikopoulos this is mainly done to identify whether a default branch of repo is `main` or `master`. There is no need for the parameterization to be handed to downstream customers. 
Do you see any use cases where this flow is much more beneficial to downstream consumers for validation of schema changes in comparison to the main or master branch?

actions/ci-beholder-validator/action.yml
anukin: Replied above.

actions/ci-beholder-validator/action.yml
anukin: Yeah this can be removed. I made it intentional to opt out of aws based docker registries if needed. I think thats over engineering and can be removed.

anukin: removed this

actions/ci-beholder-validator/action.yml
anukin: this is removed

actions/ci-beholder-validator/action.yml
anukin: The entire code will be running in runner and would be interacting with a redpanda schema registry brought through docker. Do you see any security vulnerabilities with this in that case?

erikburt: Any `${{ ... }}` used in Github Actions shell is dangerous because it's unchecked string replacement. This could technically cause arbitrary code execution.

For example, if the default branch is `main);echo "hello world"`  (brutally obvious example).

Then this  would effectively be:
```
changed_files=$(git diff --name-only origin/main);echo "hello world"
```

actions/ci-beholder-validator/action.yml
DimitriosNaikopoulos: The issue here would be if the `default_branch` was changed to a weird name for code injection. But since the default branch is always controlled by Admins even in public repos this is not as easy to take advantage of (change the branch name to something malicious). @erikburt I think you fixed some issues regarding this on the `.github` repo do you think this needs to be changed as well or because of the nature of the default_branch we are good as is?

actions/ci-beholder-validator/README.md
```diff
@@ -0,0 +1,278 @@
...
+   - If you see “pull access denied,” confirm you’re logged in or have correct
+     permissions.
+4. File Not Detected
+   - Verify your PR modifies .proto, .avsc, or a path included under
+     on.pull_request.paths.
+   - Ensure beholder.yaml is in the correct location.
+
+## Contributing
+
+Contributions are welcome! Please submit PRs with improvements or bug fixes.
```
erikburt, line 278: Is this going to be used by people external to CL? If yes, then the requirement of ECR permissions will not work.

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,107 @@
...
+    - name: Login to aws ecr
+      id: login-ecr
+      uses: aws-actions/amazon-ecr-login@062b18b96a7aff071d4dc91bc00c4c1a7945b076 # v2.0.1
+      with:
+        registries: ${{ inputs.aws-account-number }}
+
+    - name: Pull schema-validator image from ECR
+      shell: bash
+      run: |
+        docker pull ${{ steps.login-ecr.outputs.registry }}/atlas-beholder-schema-validator:${{ inputs.image-tag }}
```
erikburt, line 77:
```suggestion
    - name: Pull schema-validator image from ECR
      shell: bash
      env:
        IMAGE_TAG: ${{ inputs.image-tag }}
      run: |
        docker pull ${{ steps.login-ecr.outputs.registry }}/atlas-beholder-schema-validator:$IMAGE_TAG
```

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,107 @@
...
+      with:
+        fetch-depth: ${{ inputs.checkout-repo-fetch-depth }}
+
+    - name: Detect changed files
+      shell: bash
+      run: |
+        git fetch --prune origin +refs/heads/*:refs/remotes/origin/*
+
+        # Collect the changed files from the default branch
+        changed_files=$(git diff --name-only origin/${{ github.event.repository.default_branch }})
```
erikburt, line 44: Here as well: `${{ github.event.repository.default_branch }}`

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,107 @@
...
+          for file in $changed_files; do
+            echo "./$file"
+          done
+          echo "EOF"
+        } >> "$GITHUB_ENV"
+
+    - name: Start Redpanda with Docker Compose
+      shell: bash
+      run: |
+        docker compose -f "${{ github.action_path }}/docker-compose.yml" up -d redpanda-console
```
erikburt, line 58:
```suggestion
    - name: Start Redpanda with Docker Compose
      shell: bash
      env:
        ACTIONS_PATH: ${{ github.action_path }}
      run: |
        docker compose -f "$ACTIONS_PATH/docker-compose.yml" up -d redpanda-console
```

actions/ci-beholder-validator/action.yml
erikburt: Will also want to do this with `${{ inputs.image-tag }}`

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,107 @@
...
+      shell: bash
+      env:
+        CHANGED_FILES: ${{ env.CHANGED_FILES }}
+      run: |
+        # Validate only changed schemas
+        docker run -v "$PWD:/usr/src/app/repo" \
+          --network redpanda_network \
+          -e SCHEMA_REGISTRY_URL=http://redpanda-0:8081 \
+          -e CHANGED_FILES \
+          ${{ steps.login-ecr.outputs.registry }}/atlas-beholder-schema-validator:${{ inputs.image-tag }} \
```
erikburt, line 106: use env for `${{ inputs.image-tag}}`

erikburt (changes requested):

actions/ci-beholder-validator/action.yml
anukin: This is resolved

actions/ci-beholder-validator/action.yml
anukin: This is resolved

actions/ci-beholder-validator/action.yml
anukin: this is done

actions/ci-beholder-validator/README.md
anukin: This is removed.

actions/ci-beholder-validator/action.yml
anukin: This is resolved.

actions/ci-beholder-validator/action.yml
```diff
@@ -0,0 +1,117 @@
...
+
+runs:
+  using: composite
+  steps:
+    - name: Checkout repo
+      uses: actions/checkout@v4.2.1
+      with:
+        fetch-depth: ${{ inputs.checkout-repo-fetch-depth }}
+
+    - name: Detect changed files
```
DimitriosNaikopoulos, line 38: There is an action that is is used in different workflows that can automatically get you the changed files that we are using in a lot of repos: https://github.com/smartcontractkit/chainlink-deployments/blob/main/.github/actions/migration/build-binary/action.yaml#L37-L48 as an example

DimitriosNaikopoulos (approved):
LGTM ✅ with a minor future improvement if you want to use an external GHA to make finding of file changes easier.

erikburt (approved):

PR #831, opened 2025-01-22
Set the default port for the dynamic proxy to 443

njegosrailic:
## What 

See title. 

## Why 

The majority of workflows use 443 as the default. This doesn't conflict with anything because traffic has to be explicitly routed through it.

### DIFF
--- a/.changeset/smart-rules-hunt.md
+++ b/.changeset/smart-rules-hunt.md
@@ -1,0 +1,5 @@
+---
+"setup-gap": minor
+---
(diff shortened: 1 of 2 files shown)


chainchad (approved):
I imagine we're launching the container as root if we need it to be available on the host at 443?

njegosrailic:
> I imagine we're launching the container as root if we need it to be available on the host at 443?

I think so. Do you see any downside?

PR #840, opened 2025-01-27
[TT-xxxx] fix cache restoration errors

Tofel:
Get rid of cache restoration errors by checking toolchain version.

### DIFF
--- a/actions/ctf-setup-go/action.yml
+++ b/actions/ctf-setup-go/action.yml
@@ -40,11 +40,33 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Get Go Version
+      shell: bash
+      if: ${{ inputs.go_mod_path }}
+      id: go-version
+      run: |
+        version=$(sed -ne '/^toolchain /s/^toolchain go//p' ${{ inputs.go-go_mod_path }})
+        if [ -z "$version" ]; then
+          version=$(sed -ne '/^go /s/^go //p' ${{ inputs.go_mod_path }})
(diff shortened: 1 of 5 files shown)


actions/ctf-setup-go/action.yml
```diff
@@ -40,11 +40,33 @@ inputs:
...
+      id: go-version
+      run: |
+        version=$(sed -ne '/^toolchain /s/^toolchain go//p' ${{ inputs.go-go_mod_path }})
+        if [ -z "$version" ]; then
+          version=$(sed -ne '/^go /s/^go //p' ${{ inputs.go_mod_path }})
+          echo "Toolchain version not found in ${{ inputs.go_mod_path }}, using go directive instead."
+        fi
+        echo "Go Version: $version"
+        echo "version=$version" >> "$GITHUB_OUTPUT"
+    - name: Export Go version
```
DimitriosNaikopoulos, line 55: I came to a similar solution for the setup-go here https://github.com/smartcontractkit/.github/pull/841. 

I think you can avoid having this extra step here by doing something like this https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example on line [69](https://github.com/smartcontractkit/.github/pull/840/files#diff-5d74345f0a609429ec878b1eafef68bb2a57198a72f490e12ff9640e51273a74R69). You could check in-line if the user wants a spesific version and use that else it would use the one from go.mod

Tofel:
implemented in another PR

PR #855, opened 2025-02-04
chore: pin all actions/* references to major version

erikburt:
### Changes

Change all pinned actions/* actions to their major version tag instead.

### Motivation

Github likes to break things, and updating dependencies is a pain.


chainchad (approved):
Thanks for doing this <3

gheorghestrimtu (approved):

PR #866, opened 2025-02-06
[TT-1991] checkout repository in the same version as the chainlink_version

Tofel:
We are passing the chainlink version to test as `inputs.chainlink_version`, but whenever we checkout the calling repository (which is almost always `chainlink` or `chainlink-like`) then we checkout the latest commit from the default branch (in `chainlink` case that's `develop`). 

That might lead to weird issues, when the test is expecting that application behaves in one way (characteristic for latest commit in the default branch), but it doesn't because the image used in the test is based on `inputs.chainlink_version`.

Therefore, I think we should should be checking out repository version that corresponds to the version of application under test.

### DIFF
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -19,12 +19,6 @@ on:
           develop or commit sha"
         required: false
         type: string
-      chainlink_upgrade_version:
-        description:
-          "Enter Chainlink version to use for the upgrade tests. Example:
-          v2.10.0, develop or commit sha"
-        required: false
-        type: string
       test_path:
         description:
           "Path to the YAML test configuration file. Example:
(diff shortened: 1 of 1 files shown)


gheorghestrimtu (approved):

.github/workflows/run-e2e-tests.yml
```diff
@@ -1192,12 +1194,22 @@ jobs:
...
 
-      - name: Set short SHA
-        id: set_short_sha
+      - name: Set Slack references
+        id: set_slack_references
         shell: bash
-        run:
-          echo "short_sha=$(echo ${{ github.sha }} | cut -c1-7)" >>
-          "$GITHUB_OUTPUT"
+        run: if [ "$GITHUB_EVENT_NAME" = "pull_request" ]; then cl_ref="${{
```
lukaszcl, line 1200: @Tofel please check actionlint failures/comments

Tofel: wow, prettier did really eviscerate that one...

lukaszcl (approved):

lukaszcl (approved):

PR #869, opened 2025-02-10
feat: setup-postgres resuable action

jmank88:
Based off of: https://github.com/smartcontractkit/chainlink/tree/develop/.github/actions/setup-postgres

### Testing

https://github.com/smartcontractkit/chainlink/pull/16309

### DIFF
--- a/actions/setup-postgres/action.yml
+++ b/actions/setup-postgres/action.yml
@@ -1,0 +1,18 @@
+name: Setup Postgresql
+description: Setup postgres docker container via docker-compose, allowing usage of a custom command, see https://github.com/orgs/community/discussions/26688
+inputs:
+  base-path:
+    description: Path to the base of the repo
+    required: false
+    default: .
+runs:
+  using: composite
+  steps:
+    - name: Start postgres service
+      run: docker compose up -d
+      shell: bash
+      working-directory: ${{ inputs.base-path }}/.github/actions/setup-postgres
+    - name: Wait for postgres service to be healthy
+      run: ./wait-for-healthy-postgres.sh
+      shell: bash
+      working-directory: ${{ inputs.base-path }}/.github/actions/setup-postgres
--- a/actions/setup-postgres/action.yml
+++ b/actions/setup-postgres/action.yml
@@ -1,18 +1,51 @@
 name: Setup Postgresql
-description: Setup postgres docker container via docker-compose, allowing usage of a custom command, see https://github.com/orgs/community/discussions/26688
+description:
+  Setup postgres docker container via docker-compose, allowing usage of a custom
+  command, see https://github.com/orgs/community/discussions/26688
 inputs:
-  base-path:
-    description: Path to the base of the repo
+  print-logs:
+    description: |
+      Whether to print the logs of the postgres service to the console and a file.
+      Note: This action should be called first without this input set.
     required: false
-    default: .
+    default: ""
+  print-logs-path:
+    description: |
+      The directory to print the logs to.
+      Note: This input does nothing if `print-logs` is not set to true.
+    default: ${{ github.workspace }}
+
+outputs:
+  database-url:
+    description: The URL to connect to the postgres database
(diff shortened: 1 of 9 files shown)


actions/setup-postgres/action.yml
```diff
@@ -0,0 +1,18 @@
...
+  using: composite
+  steps:
+    - name: Start postgres service
+      run: docker compose up -d
+      shell: bash
+      working-directory: ${{ inputs.base-path }}/.github/actions/setup-postgres
+    - name: Wait for postgres service to be healthy
+      run: ./wait-for-healthy-postgres.sh
+      shell: bash
+      working-directory: ${{ inputs.base-path }}/.github/actions/setup-postgres
```
erikburt, line 18: I think this is the correct way to accomplish what you want. The action's path should be populated by Github and then it should have access to the files in this action's directory (like .env, docker-compose.yml, wait-for-healthy-postgres).

```suggestion
inputs:
runs:
  using: composite
  steps:
    - name: Start postgres service
      run: docker compose up -d
      shell: bash
      working-directory: ${{ env.GITHUB_ACTION_PATH }}
    - name: Wait for postgres service to be healthy
      run: ./wait-for-healthy-postgres.sh
      shell: bash
      working-directory: ${{ env.GITHUB_ACTION_PATH }}
```

actions/setup-postgres/bin/pg_dump
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+#
+# This script acts as a docker replacement around pg_dump so that developers can
+# run DB involved tests locally without having postgres installed.
+
+cd "$(dirname "$0")" || exit
+
+docker compose exec -T postgres pg_dump "$@"
```
erikburt, line 8: This file is not referenced by this action

erikburt (changes requested):

actions/setup-postgres/bin/pg_dump
jmank88: Any idea why it is in the core repo?

erikburt: I think it used to be used in ci-core for debugging purposes.

jmank88:
In use here:
- https://github.com/smartcontractkit/chainlink-integrations/pull/7
- https://github.com/smartcontractkit/chainlink/pull/16311

Seems to work fine :shrug:

actions/setup-postgres/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+name: gha_postgres
+services:
+  postgres:
+    ports:
+      - "5432:5432"
+    container_name: cl_pg
+    image: postgres:14-alpine
```
pavel-raykov, line 7: is the postgres version meant to be customizable? E.g., would it be possible to compatible with https://github.com/smartcontractkit/chainlink/blob/develop/.tool-versions ?

erikburt: This is mostly copying the functionality needed by ci-core. I can add a task to releng's queue to make this configurable?

Is this something that is needed sooner rather than later?

pavel-raykov (approved):

actions/setup-postgres/docker-compose.yml
jmank88: We might have separate concerns in CI thought. Typically we run the oldest supported version to ensure that we haven't broken compatability. `.tool-versions` on the other hand might be on a newer version

njegosrailic (approved):

actions/setup-postgres/docker-compose.yml
pavel-raykov: This is actually a very appealing use case - ideally, we could run CI on a matrix of postgres versions to ensure that we work correctly with all the versions (or a major subset) that we claim to support.

chainchad (approved):

erikburt (approved):

PR #882, opened 2025-02-13
Improving GAP GHA configuration in the ctf-setup-run-tests-environment GHA

njegosrailic:
## What 

See title. 

## Why 

Not all tests need GAP and access to internal services. Also, we need a flag to disable GAP, as this action is being used for different tests and environments.

### DIFF
--- a/actions/ctf-setup-run-tests-environment/action.yml
+++ b/actions/ctf-setup-run-tests-environment/action.yml
@@ -61,19 +61,29 @@ inputs:
   gati_token:
     required: false
     description: Token provided by GATI to pull from private repos
+  enable-gap:
+    required: false
+    default: "true"
+    description: Configure and enable GAP for accessing internal services.
   main-dns-zone:
-    required: true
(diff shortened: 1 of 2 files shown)


actions/ctf-setup-run-tests-environment/action.yml
```diff
@@ -61,19 +61,29 @@ inputs:
...
 
 runs:
   using: composite
   steps:
+    - name: Validate GAP inputs
+      shell: bash
+      if: inputs.enable-gap == 'true'
+      run: |
+        if [ -z "${{ inputs.main-dns-zone }}" ]; then
+          echo "Error: main-dns-zone is required when enable-gap is true."
```
Tofel, line 83: shouldn't we validate `k8s-cluster-name` as well?

njegosrailic: Thanks.

rafaelfelix (approved):

PR #899, opened 2025-02-20
[INFOPLAT-1914] Beholder-Pulumi: Preview stage and state lock handling

nanchano:
Fixes state lock bug during runtime.

Adds preview stage.


pkcll (approved):

actions/beholder-pulumi-deploy-schema/action.yml
```diff
@@ -57,14 +57,46 @@ runs:
...
+          -e DEPLOY_CONFIG_PATH=${{ inputs.deploy_config_path }} \
+          -e PULUMI_S3_BUCKET=${{ inputs.pulumi_s3_bucket }} \
+          -e CONFLUENT_CLOUD_API_KEY=${CONFLUENT_CLOUD_API_KEY} \
+          -e CONFLUENT_CLOUD_API_SECRET=${CONFLUENT_CLOUD_API_SECRET} \
+          -e SCHEMA_REGISTRY_KEY=${SCHEMA_REGISTRY_KEY} \
+          -e SCHEMA_REGISTRY_SECRET=${SCHEMA_REGISTRY_SECRET} \
+          -e PULUMI_CONFIG_PASSPHRASE=${PULUMI_CONFIG_PASSPHRASE} \
+          -e KAFKA_API_KEY=${KAFKA_API_KEY} \
+          -e KAFKA_API_SECRET=${KAFKA_API_SECRET} \
+          -e KAFKA_REST_ENDPOINT=${KAFKA_REST_ENDPOINT} \
```
chainchad, line 93: Maybe as a follow-up PR, but I'd probably put all of the env vars that are required here as inputs to this action instead of env vars. Then you'd pass the inputs in via the `env: {}` object (like you're doing below but the value would be the `${{ input.<key> }}`. 

This will ensure that the caller knows all of the inputs that the action needs. If you pass in secrets via `inputs` they will remain as a secret (masked).

chainchad (approved):

actions/beholder-pulumi-deploy-schema/action.yml
nanchano: will do in a follow up PR when we refactor credentials, ty

PR #902, opened 2025-02-20
Create composite workflow wrapper for dependency-review

chainchad:


actions/dependency-review/action.yml
```diff
@@ -0,0 +1,76 @@
...
+      env:
+        CONFIG_FILE: ${{ inputs.config-file }}
+        CONFIG_PRESET: ${{ inputs.config-preset }}
+      run: |
+        if [[ -n "${CONFIG_FILE:-}" ]]; then
+          echo "Using config file: ${CONFIG_FILE}"
+          echo "path=${CONFIG_FILE}" | tee -a "$GITHUB_OUTPUT}"
+        elif [[ -n "${CONFIG_PRESET:-}" ]]; then
+          echo "Using config preset: ${CONFIG_PRESET}"
+          echo "path=${GITHUB_ACTION_PATH}/configs/${CONFIG_PRESET}.yml" | tee -a "$GITHUB_OUTPUT}"
```
DimitriosNaikopoulos, line 70: Just out of curiosity what is the advantage of using `tee` instead of `>>` when writing a string in the outputs? 🤔

actions/dependency-review/action.yml
```diff
@@ -0,0 +1,76 @@
...
+    - name: Input validation
+      shell: bash
+      env:
+        CONFIG_FILE: ${{ inputs.config-file }}
+        CONFIG_PRESET: ${{ inputs.config-preset }}
+      run: |
+        if [[ -z "${CONFIG_FILE:-}" && -z "${CONFIG_PRESET:-}" ]]; then
+          echo "::error::Either 'config-preset' or 'config-file' inputs must be provided."
+          exit 1
+        elif [[ -n "${CONFIG_FILE:-}" && -n "${CONFIG_PRESET:-}" ]]; then
```
DimitriosNaikopoulos, line 51: Should one of the 2 options has a default value with one of the `configs` available here? Then if someone wants to use it with the default option they can just add this action in a workflows and it should work since this repo is public and I think we could read the configs without a GATI 🤔

DimitriosNaikopoulos (approved):
LGTM ✅ Just 1 question and 1 suggestion 😄

actions/dependency-review/action.yml
chainchad: Good question, `>>` will redirect all output to the env var with nothing printed to stdout. Piping to `tee` will send to stdout AND to this env var. Adding the `-a` flag to tee will 'append' which is important because otherwise it will replace the existing values of that env var.

tl;dr - saves me an echo for debugging the output :)

actions/dependency-review/action.yml
chainchad: Discussed out-of-band. Going to leave as-is for now to keep things moving but we should look at this in the future, perhaps before making this an org-wide check.

PR #911, opened 2025-02-28
feat: update crib deploy env - ns override

ajgrande924:
- create `ns-name-override` input
- do not error out if ns exists
- override labels

### DIFF
--- a/actions/crib-deploy-environment/action.yml
+++ b/actions/crib-deploy-environment/action.yml
@@ -166,7 +166,7 @@ runs:
         else
           echo "devspace-namespace=${NS_NAME_OVERRIDE}" | tee -a "$GITHUB_OUTPUT"
         fi
-        
+
         # configure rest of variables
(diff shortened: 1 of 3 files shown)


actions/crib-deploy-environment/action.yml
```diff
@@ -257,9 +269,9 @@ runs:
         fi
 
         echo "Creating ${DEVSPACE_NAMESPACE} Kubernetes namespace.."
-        kubectl create ns "${DEVSPACE_NAMESPACE}"
+        kubectl get ns "${DEVSPACE_NAMESPACE}" || kubectl create ns "${DEVSPACE_NAMESPACE}"
```
njegosrailic, line 272: The only thing that worries me is that maybe we should have a flag if we want to enable reuse of a namespace/override labels. Recently, we saw conflicts, such as the same UUID being generated multiple times. 

Anyway, it is gtg thanks for the improvements. Can you please just add a change set?

njegosrailic (approved):

PR #941, opened 2025-03-18
Set CRIB eks cluster name

scheibinger:


Tofel (approved):
I recommend this PR to everyone that looks to buy a PR. It does everything it advertises and even better than that!

erikburt (approved):

PR #971, opened 2025-03-26
Flakeguard AI environment variable adjustment and file path fix

KMontag42:
ensures that we are using the `/github/workspace` directory for file lookup.
forwards the file name correctly from the workflow supplied input.

### DIFF
--- a/actions/flakeguard-ai-analysis/main.py
+++ b/actions/flakeguard-ai-analysis/main.py
@@ -22,6 +22,7 @@ def main():
     )
     analyzer = FlakyTestAnalyzer(test_guide=test_guide)
     failed_test_results = os.environ.get("FAILED_TEST_RESULTS", "context/results.json")
+    logger.info(f"Checking for results file at {file_path / failed_test_results}")
     if not Path(file_path / failed_test_results).exists():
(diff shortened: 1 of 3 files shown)


actions/flakeguard-ai-analysis/main.py
```diff
@@ -12,7 +12,7 @@
 
 
 def main():
-    file_path = Path(__file__).parent
+    file_path = Path("/github/workspace")
```
erikburt, line 15: Should this be a configurable input?

Also `$GITHUB_WORKSPACE` is an env that is also available, not sure if it's the directory you need though.

Slightly more information here: https://github.com/actions/checkout/issues/785

erikburt (approved):

actions/flakeguard-ai-analysis/main.py
KMontag42: so because this is running within a docker container, the mount path is always the same.

example:
```/usr/bin/docker run ... -v "/home/runner/work/chainlink/chainlink":"/github/workspace" ...```

from https://github.com/smartcontractkit/chainlink/actions/runs/14087667410/job/39461771982

PR #974, opened 2025-03-26
Support github token as a mounted secret in docker build

chainchad:


actions/build-push-docker/action.yml
```diff
@@ -197,6 +201,8 @@ runs:
           ${{
             format('{0},scope={1}', inputs.docker-build-cache-to, runner.arch)
           }}
+        secrets: |
+          GIT_AUTH_TOKEN=${{ inputs.github-token }}
```
erikburt, line 205: This allows it to pull go code from private repositories during a build?

erikburt (approved):

actions/build-push-docker/action.yml
chainchad: Yeah, exactly. The Dockerfile will still need to mount the secret for it to be usable though.

PR #977, opened 2025-03-28
Update inputs to be read from env vars explicitly

chainchad:


erikburt (approved):

actions/ci-benchmarking/action.yml
```diff
@@ -61,14 +61,18 @@ runs:
         go-version: "stable"
 
     - name: Analyze GitHub Event
+      env:
+        GITHUB_EVENT_NAME: ${{ github.event_name }}
```
patrickhuie19, line 65: I'm not sure I understand the remediation - since the templated value is filled in before the workflow executes, won't all values of `${{ github.event_name }}` be expanded the same?

chainchad: Discussing in a separate thread.

patrickhuie19 (approved):

erikburt (approved):

PR #1023, opened 2025-05-02
Create docker-image-patch action

chainchad:
This creates a Docker container from a specified source image and:

1. Copies in a list of files to their specified destination for e.g. plugins.
2. Creates and sets a specified list of ENV directives.
3. Creates a new image and publishes it to the specified ECR destination.


actions/docker-image-patch/action.yml
```diff
@@ -0,0 +1,192 @@
...
+        COPY_PATHS: ${{ inputs.copy-paths }}
+        DOCKER_ENVS: ${{ inputs.docker-envs }}
+        DOCKER_REGISTRY_URL_DST: ${{ inputs.docker-registry-url-dst }}
+        DOCKER_REGISTRY_URL_SRC: ${{ inputs.docker-registry-url-src }}
+      run: |
+        echo "::group::Validate inputs"
+        ${GITHUB_ACTION_PATH}/scripts/input-validation.sh
+        echo "::endgroup::"
+
+    - name: Set outputs
```
erikburt, line 108: Nit -

It seems unintuitive to "set outputs" before everything else is completed. I see that these values are used in subsequent steps, and I also see that this action has no outputs declared.

So maybe this step could be called 'process-inputs' or something?

erikburt (approved):

actions/docker-image-patch/action.yml
chainchad: Yeah, it's confusing because there are outputs at the step level, job level, and also the composite action level.

PR #1029, opened 2025-05-12
Create reusable workflow for dependency-review

chainchad:
Needs:
- [x] https://github.com/smartcontractkit/.github/pull/1028 merged
- [x] A new release for dependency-review composite action.


.github/workflows/reusable-dependency-review.yml
```diff
@@ -0,0 +1,42 @@
...
+      contents: read
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Dependency Review
+        uses: smartcontractkit/.github/actions/dependency-review@dependency-review/2.0.0
```
erikburt, line 40: Is it worth using `/v2` tag here? Meaning we wouldn't have to update this workflow, and all references to it if the action is updated?

erikburt (approved):

.github/workflows/reusable-dependency-review.yml
chainchad: Yeah, definitely. Good thoughts. Forgot you added that now.

Updated in 890ff3fb.

erikburt (approved):

PR #1033, opened 2025-05-13
Cld 205/ignore upload on merge queue

bytesizedroll:
## Feat: Add `skip-cache-save` option to setup-golang

This change adds a new `skip-cache-save` input (defaults to `false`).

If you set `skip-cache-save: true`:
* The action will still try to **download (restore)** an existing cache.
* The action will **not upload (save)** a new cache.

This is useful to speed up workflows when you don't need to update the cache.


actions/setup-golang/action.yml
```diff
@@ -21,6 +21,11 @@ inputs:
     description: Set this to cache bust
     required: false
     default: "1"
+  skip-cache-save:
+    description:
+      Set to 'true' to skip saving the cache while still allowing restores
+    required: false
+    default: "false"
```
erikburt, line 28: The cache is saved at the end of the workflow, adding `actions/cache/save@v4` here will save it immediately, which I assume is not the goal.

The proper way to do this is to call `actions/cache/restore@v4` when you don't want to save it, and `actions/cache@v4` when you do want to save it.

erikburt (changes requested):

actions/setup-golang/action.yml
bytesizedroll: ahh yes, thanks! didnt realize that save wont also create that post-workflow upload

erikburt (approved):

PR #1062, opened 2025-05-29
feat: run-e2e-tests optionally setup db

erikburt:
### Changes

- Don't setup the DB at the job level for `in-memory-tests` job
  -  See related PRs: #1060, #1061 
- Don't cache `go` for `load-test-configurations` job
  - This routinely takes a very long time, not sure if it's necessary for such a simple tool like ci-tool.
  - Example: 5min setup-go step, and 5min post setup-go step - https://github.com/smartcontractkit/chainlink/actions/runs/15334488256/job/43148823751?pr=17943#step:3:31 
- Don't install jq as it comes automatically installed on runners

### Testing

See https://github.com/smartcontractkit/chainlink/pull/17943

---

DX-931 + DX-934

### DIFF
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -838,7 +838,7 @@ jobs:
       - name: Run tests
         id: run_tests
         timeout-minutes: ${{ fromJson(env.DOCKER_TESTS_TIMEOUT_MINUTES) }}
-        uses: smartcontractkit/.github/actions/ctf-run-tests@feat/speed-up-in-memory-tests-run-tests
+        uses: smartcontractkit/.github/actions/ctf-run-tests@ctf-run-tests/0.8.0
         env:
           DETACH_RUNNER: true
           E2E_TEST_CHAINLINK_VERSION:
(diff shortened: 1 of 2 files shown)


.github/workflows/run-e2e-tests.yml
```diff
@@ -445,17 +441,18 @@ jobs:
         uses: actions/checkout@v4.2.1
         with:
           persist-credentials: false
+
       - name: Setup Go
         uses: actions/setup-go@v5.0.2
         with:
           go-version: "1.24.0"
           check-latest: true
+          cache: false # disable caching as this job doesn't benefit from it
```
kalverra, line 450: Do we have a test run to check this? Is it not using runs-on?

kalverra (changes requested):

.github/workflows/run-e2e-tests.yml
kalverra: I'm just surprised this is the case.

erikburt: > Is it not using runs-on?

These tests are not using runs-on yet. A job like this would also not be migrated to self-hosted runners due to it's simplicity. This is only generating the test matrix JSON.

Migrating it to SHRs would incur a ~20s cost to boot the runner, rather than the faster boot/queue times when using a GH runner like `ubuntu-latest`.

kalverra (approved):

PR #1119, opened 2025-06-25
Bump ctf-run-tests version in run-e2e-tests workflow

chudilka1:


actions/parse-and-mask-test-secrets/action.yml
```diff
@@ -0,0 +1,24 @@
...
+  using: composite
+  steps: 
+    - name: Checkout code
+      uses: actions/checkout@v4.2.1
+
+    - name: Setup Go
+      uses: actions/setup-go@v5.0.2
+      with:
+        go-version: "1.24.0"
+        cache: true
```
erikburt, line 19: Do not cache here. It is likely not worth it.

actions/parse-and-mask-test-secrets/action.yml
```diff
@@ -0,0 +1,24 @@
...
+inputs:
+  encoded_test_secrets:
+    required: true
+    description: The base64-encoded test secrets file
+
+runs:
+  using: composite
+  steps: 
+    - name: Checkout code
+      uses: actions/checkout@v4.2.1
```
erikburt, line 13: This is not needed right?

actions/parse-and-mask-test-secrets/mask-testsecrets/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/smartcontractkit/chainlink-github-actions/chainlink-testing-framework/run-tests/mask-testsecrets
+
+go 1.21.3
```
erikburt, line 3: `setup-go` should probably use this go version.

.github/workflows/run-e2e-tests.yml
```diff
@@ -421,13 +421,17 @@ jobs:
     name: Load test configurations
```
erikburt, line 421: This doesn't seem like the place for this added logic.

.github/workflows/run-e2e-tests.yml
```diff
@@ -610,8 +643,9 @@ jobs:
     name: Get Chainlink image
     needs: [validate-inputs, load-test-configurations]
     if: ${{
```
erikburt, line 645: Can you perform the above logic here, and then skip the image build if its a public ECR, rather than throwing the logic into `load-test-configurations`. That or make a separate job completely.

erikburt (changes requested):

.github/workflows/run-e2e-tests.yml
```diff
@@ -600,6 +604,35 @@ jobs:
...
+        uses: smartcontractkit/.github/actions/ctf-fetch-aws-secret@921f4b0ca850dd473dcef9082e3169ccbb83cc52 # ctf-fetch-aws-secret@0.0.0
+        with:
+          secret_id: ${{ needs.validate-inputs.outputs.aws_test_secrets_key }}
+          aws_region: ${{ secrets.QA_AWS_REGION }}
+          aws_role_to_assume: ${{ secrets.QA_AWS_ROLE_TO_ASSUME }}
+
+      - name: Parse and Mask Test Secrets
+        if: ${{ needs.validate-inputs.outputs.aws_test_secrets_key }}
+        id: parsed-aws-test-secrets
+        uses: smartcontractkit/.github/actions/parse-and-mask-test-secrets@e2a70340b9a4feb6a6724dc2add74101ed425e70
```
erikburt, line 619: Once the action is added to develop, you can add the proper tag here.

actions/parse-and-mask-test-secrets/mask-testsecrets/go.mod
chudilka1: Thanks, done

actions/parse-and-mask-test-secrets/action.yml
chudilka1: yes, thanks, removed

.github/workflows/run-e2e-tests.yml
```diff
@@ -605,13 +605,54 @@ jobs:
...
+        id: aws-test-secrets
+        uses: smartcontractkit/.github/actions/ctf-fetch-aws-secret@921f4b0ca850dd473dcef9082e3169ccbb83cc52 # ctf-fetch-aws-secret@0.0.0
+        with:
+          secret_id: ${{ needs.validate-inputs.outputs.aws_test_secrets_key }}
+          aws_region: ${{ secrets.QA_AWS_REGION }}
+          aws_role_to_assume: ${{ secrets.QA_AWS_ROLE_TO_ASSUME }}
+
+      - name: Parse and mask test secrets
+        id: parsed-aws-test-secrets
+        uses: smartcontractkit/.github/actions/parse-and-mask-test-secrets/0.1.0 # parse-and-mask-test-secrets@0.1.0
```
erikburt, line 631: `uses: smartcontractkit/.github/actions/parse-and-mask-test-secrets@parse-and-mask-test-secrets/0.1.0`

.github/workflows/run-e2e-tests.yml
```diff
@@ -605,13 +605,54 @@ jobs:
...
   # Check if Chainlink images required for the tests exist. If not, build and push the images to QA ECR
+  # Skip if public registry used in test secrets or image build is explicitly skipped
   require-chainlink-image-versions-in-qa-ecr:
     name: Get Chainlink image
-    needs: [validate-inputs, load-test-configurations]
+    needs: [validate-inputs, load-test-configurations, check-registry-type]
     if: ${{
-      fromJson(needs.validate-inputs.outputs.require_chainlink_image_versions_in_qa_ecr_matrix)
-      != '' && !inputs.skip_image_build }}
+      fromJson(needs.validate-inputs.outputs.require_chainlink_image_versions_in_qa_ecr_matrix) != ''
```
erikburt, line 653: I'm kind of confused as to what's happening here. The image this is configured to build is going to `${{ secrets.QA_AWS_ACCOUNT_NUMBER }}.dkr.ecr.${{ secrets.QA_AWS_REGION }}.amazonaws.com/chainlink` ECR. But the above checks if it's going to use a public ECR and skips this image if so.

This seems like an indirect way of skipping the image build based on whats in some value in AWS, which doesn't even seem to be a secret, because it's literally a public ECR.

- Why is the image being injected as a 'test secret' from AWS?

See build chainlink image step below - not configured to build or publish to a public ECR.

```
      - name: Build Chainlink Image
        if: ${{ steps.check-image-exists.outputs.exists != 'true' }}
        uses: smartcontractkit/.github/actions/ctf-build-image@ctf-build-image/v1
        with:
...
          docker-registry-url: ${{ secrets.QA_AWS_ACCOUNT_NUMBER }}.dkr.ecr.${{ secrets.QA_AWS_REGION }}.amazonaws.com
          docker-repository-name: "chainlink"
...
```

erikburt: I understand that run-e2e-tests is a complete mess and needs to be rewritten, but I'm failing to see what this change is actually trying to do.

You can skip the image build with `skip_image_build` input.

But you'll only know if you need to skip the image build if the value stored in an AWS secret is in the public ECR?

So then as part of the calling workflow, couldn't you get the test 'secrets', check the image there, and then have that effect the input of `skip_image_build`, instead of putting this single use-case logic inside a reusable workflow?

erikburt: For example, move the `check-registry-type` type job into [your workflow](https://github.com/smartcontractkit/chainlink/blob/55af3732f1c262bf2d8bfd567e1c1ef1b91ef399/.github/workflows/on-demand-ocr-soak-test.yml#L43). 

Then that job outputs `skip_image_build`, then when you call `run-e2e-test.yml` pass the input:

```
check-registry-type:
  outputs:
    skip_image_build: ${{ steps. ... .outputs.is-public-registry }}
  ...

run-e2e-tests-workflow:
  needs: check-registry-type
  uses: ...run-e2e-tests.yml@...
  with:
    ...
    skip_image_build: ${{ needs.check-registry-type.outputs.skip_image_build }}
```

chudilka1: Good point @erikburt, thanks, updated.

erikburt (approved):

PR #1120, opened 2025-06-26
Add parse-and-mask-test-secrets action

chudilka1:


actions/ctf-run-tests/action.yml
```diff
@@ -323,18 +323,16 @@ runs:
...
     # Custom secrets override the default secrets
     - name: Set custom test secrets from dotenv file
       if: inputs.test_secrets_override_base64 != ''
-      shell: bash
-      run:
-        go run ${{ github.action_path }}/mask-testsecrets/main.go "${{
-        inputs.test_secrets_override_base64 }}"
+      uses: smartcontractkit/.github/actions/parse-and-mask-test-secrets@v0.0.1 # parse-and-mask-test-secrets@0.0.1
+      with:
+        encoded_test_secrets: ${{ inputs.test_secrets_override_base64 }}
```
erikburt, line 335: This should not be part of this PR. These changes should be made afterwards.

Please leave `ctf-run-tests` as-is, and then when you migrate over to the new action, delete all redundant sources.

actions/parse-and-mask-test-secrets/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "parse-and-mask-test-secrets",
+  "version": "0.0.1",
```
erikburt, line 3: Leave this as `0.0.0`. This  will be bumped and tagged by `changesets`

Run `pnpm changeset add` to add a changeset for this action.

erikburt (changes requested):

actions/ctf-run-tests/action.yml
chudilka1: Thanks, updated

erikburt (approved):

PR #1145, opened 2025-07-17
feat: build-push-docker optional inputs

erikburt:
### Changes

Allows for many inputs to be optional when `docker-push` is false.

### Motivation

Some users of the `ctf-build-image` action do not push the image, they are simply building the image as a sanity check.
- On second thought, why does this matter? They can build the app without building a docker image, that should accomplish something similar.


erikburt:
duplicate of #1154

PR #1174, opened 2025-08-25
Support manifest annotations on reusable workflow

chainchad:
Needs https://github.com/smartcontractkit/.github/pull/1173 merged/released.

### DIFF
--- a/.github/workflows/reusable-docker-build-publish.yml
+++ b/.github/workflows/reusable-docker-build-publish.yml
@@ -579,75 +579,62 @@ jobs:
           DOCKER_MANIFEST_TAG: ${{ needs.build-publish.outputs.docker-manifest-tag }}
           AWS_ECR_NAME: ${{ inputs.aws-ecr-name }}
         run: |
-          # Initialize arrays
-          ANNOTATIONS=()
-          USER_ANNOTATION_KEYS=()
+          # Use associative array to store the annotations.
+          declare -A annotations
 
-          # Parse user annotations once to build key lookup and add them to final array.
+          # Set the default annotations - can be overridden.
+          echo "Adding default annotations..."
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-docker-build-publish.yml
```diff
@@ -536,6 +570,83 @@ jobs:
...
+          # Add automatic annotations only if user didn't provide them.
+          for annotation in "${automatic_annotations[@]}"; do
+            key="${annotation%%=*}"
+            if ! has_user_annotation "$key"; then
+              ANNOTATIONS+=("$annotation")
+              echo "  Added automatic annotation: $annotation"
+            else
+              echo "  Skipped automatic annotation (user override): $key"
+            fi
+          done
```
erikburt, line 629: I feel like this logic is fragile and overly complex, because:
1. It allows users to supply duplicate annotations
2. It complicates the idea of 'default' annotations

Here's an example from chatgpt:

```bash
#!/usr/bin/env bash
declare -A annotations

# Defaults with dots in keys
annotations=(
  ["app.version"]="1.0"
  ["app.author"]="unknown"
  ["build.date"]="2025-08-25"
)

# Function to add or overwrite annotations from input
add_annotation() {
    local key="$1"
    local value="$2"
    annotations["$key"]="$value"
}

# Example: Process user-supplied arguments like key=value
for arg in "$@"; do
    if [[ "$arg" == *"="* ]]; then
        key="${arg%%=*}"
        value="${arg#*=}"
        add_annotation "$key" "$value"
    fi
done

# Print
for key in "${!annotations[@]}"; do
    echo "$key = ${annotations[$key]}"
done
```

chainchad: I like the associative array approach. Will give that a shot.

chainchad: 86d37817

PR #1177, opened 2025-08-25
Use latest version of Docker buildx for manifest list creation

chainchad:
- **Add setup buildx step to support --annotate flag on manifest list creation**
- **Temp: use unreleased action**

### DIFF
--- a/actions/build-push-docker-manifest/action.yml
+++ b/actions/build-push-docker-manifest/action.yml
@@ -135,6 +135,11 @@ outputs:
 runs:
   using: composite
   steps:
+    - name: Set up Docker Buildx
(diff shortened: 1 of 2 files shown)


actions/build-push-docker-manifest/action.yml
```diff
@@ -174,18 +179,15 @@ runs:
...
+                ANNOTATION_FLAGS+=("--annotation" "index:$line")
+                echo "  Adding annotation: index:$line"
               else
                 echo "  WARNING: Skipping invalid annotation format: $line"
               fi
             fi
           done <<< "$MANIFEST_ANNOTATIONS"
         fi
 
-        # Always add created timestamp
```
chainchad, line 186: This happens from the caller (reusable workflow).

PR #1245, opened 2025-09-29
fix(setup-golang): use path from input

graham-chainlink:
In multiple places, we assume the go.mod and go.sum files are located on the root which is not the case when the project is a monorepo, we should use the path provided in the input instead of assuming

Example [here](https://github.com/smartcontractkit/chainlink-deployments/actions/runs/18052817312/job/51377775501#step:8:533)

> Cache restored successfully
Cache restored from key: Linux-gobuild-1--main

Note the `--`, blank because go.sum is not found on the root

### DIFF
--- a/actions/setup-golang/action.yml
+++ b/actions/setup-golang/action.yml
@@ -76,7 +76,7 @@ runs:
           ${{ steps.go-cache-dir.outputs.gomodcache }}
         key:
           ${{ runner.os }}-gomod-${{ inputs.cache-version }}-${{
-          hashFiles('./go.sum') }}
+          hashFiles(inputs.go-cache-dep-path) }}
         restore-keys: |
           ${{ runner.os }}-gomod-${{ inputs.cache-version }}-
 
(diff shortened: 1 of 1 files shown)


gustavogama-cll (approved):

actions/setup-golang/action.yml
```diff
@@ -76,7 +76,7 @@ runs:
           ${{ steps.go-cache-dir.outputs.gomodcache }}
         key:
           ${{ runner.os }}-gomod-${{ inputs.cache-version }}-${{
-          hashFiles('./go.sum') }}
+          hashFiles(inputs.go-cache-dep-path) }}
```
gustavogama-cll, line 79: don't you need to update the default `go-cache-dep-path` to `./go.sum`?

gustavogama-cll (approved):

actions/setup-golang/action.yml
graham-chainlink: looking at the documentation [here](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#hashfiles)

It confirms that hashFiles() paths are relative to the GITHUB_WORKSPACE directory (which is the root of your repository). 
This means:
hashFiles('go.sum') → looks for go.sum in the repository root
hashFiles('./go.sum') → also looks for go.sum in the repository root

PR #1253, opened 2025-09-30
feat: branch-out-upload variant

erikburt:
### Changes

* Adds `trunk-variant` input parameter
* Small updates to docs

### DIFF
--- a/.changeset/friendly-paws-switch.md
+++ b/.changeset/friendly-paws-switch.md
@@ -1,7 +1,0 @@
----
-"ci-publish-npm": major
----
-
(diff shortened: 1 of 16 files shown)


actions/branch-out-upload/action.yml
```diff
@@ -29,13 +29,21 @@ inputs:
     default: "false"
 
   trunk-job-url:
-    description: "The URL to the job run"
+    description: The URL to the job run"
```
kalverra, line 32: Why remove quote?

kalverra (changes requested):

PR #1447, opened 2026-03-06
Medic

patrickhuie19:
Bringing https://github.com/smartcontractkit/medic-test/pulls to this repo. Next step is to use the reusable workflows in core.


.github/workflows/medic-merge-conflict.yml
```diff
@@ -0,0 +1,334 @@
...
+
+      - name: Parse comment
+        id: parse
+        run: node _medic/.github/actions/medic/dist/comment-parser/index.js
+        env:
+          INPUT_GITHUB-TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          INPUT_COMMENT-BODY: ${{ github.event.comment.body }}
+          INPUT_ISSUE-NUMBER: ${{ github.event.issue.number }}
+          INPUT_COMMENTER-LOGIN: ${{ github.event.comment.user.login }}
+
```
erikburt: This should not be needed. If the proper `dist/*/index.js` and `action.yml` is available, then you don't need need to setup node or call it like this.

You should be able to call this like:

```yml`
name: Parse comment
id: parse
uses: smartcontractkit/.github/actions/medic/comment-parser@<tag>`
```

and the `action.yml` file should default all these inputs.

This issue appears a few times in this PR.

.github/actions/medic/src/workflow-retry/github-service.ts
```diff
@@ -0,0 +1,100 @@
...
+    run_id: runId,
+    filter: 'latest'
+  });
+
+  const failedJobs: FailedJob[] = [];
+
+  for (const job of jobs) {
+    if (job.conclusion !== 'failure') continue;
+
+    const failedStep = job.steps?.find(s => s.conclusion === 'failure');
```
erikburt: Not sure if relevant, but multiple steps can end in a failure. For example, a step may always run even if there were previous failures: `if: always()`.

.github/actions/medic/src/workflow-retry/github-service.ts
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Get all jobs with conclusion: failure for a workflow run
+ */
+export async function getFailedJobs(
+  octokit: OctokitClient,
+  owner: string,
+  repo: string,
+  runId: number
+): Promise<FailedJob[]> {
```
erikburt: I would define the return typess of functions in the file they're declared. It makes it easier to read IMO. Dedicated types files (usually `types.d.ts`) are usually for generic helpers.

.github/actions/medic/src/workflow-retry/claude-client.ts
```diff
@@ -0,0 +1,333 @@
...
+): Promise<AnalysisResult | null> {
+  const projectId = process.env.ANTHROPIC_VERTEX_PROJECT_ID;
+  if (!projectId) {
+    core.info('Claude analysis not available (no Vertex credentials), using unconditional retry');
+    return null;
+  }
+
+  let client: { messages: { create: (params: Record<string, unknown>) => Promise<ApiMessage> } };
+  try {
+    const mod = await import('@anthropic-ai/vertex-sdk');
```
erikburt: Any particular reason to import this inline?

erikburt (changes requested):
I’ve left more detailed comments inline, but at this size it’s difficult to fully vet in a single pass.

Some higher-level feedback:
- I think this probably belongs in its own repo. It’s not just a single action, it’s a collection of related actions and workflows, and because of that, seems like a suite/product in itself. Having it's own repository would make ownership and versioning a lot clearer. Can expand on this further if needed.
    - I’m happy to help set up a new repo.
    - If it must live in this repository, then all the code should live in the top-level `apps/medic` directory probably, rather than `.github/actions/medic`.
- You're most of the way there, but each major medic feature should likely be its own independently versioned action, with shared functionality pulled into a common library.
    - For example comment-parser, doesn't have an action.yml but is does have it's own `dist`. `lib` is not an actual lib, it's a shared folder. Although, probably not big enough to warrant it being a separate library yet, still helps with organization imo.
- This kind of goes with the above, but right now, the code is tough to reason about and maintain. I think it would benefit from more structural cleanup. Clearer boundaries between individual functionality, tests utilities, common functions, etc.
- Documentation would be helpful as well

.github/workflows/medic-workflow-retry.yml
```diff
@@ -0,0 +1,110 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Checkout caller repo
+        uses: actions/checkout@v4
+
+      - name: Checkout medic
+        uses: actions/checkout@v4
+        with:
+          repository: smartcontractkit/medic-test
```
timweri: Do we need this?

.github/workflows/medic-merge-conflict.yml
```diff
@@ -0,0 +1,334 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Use specified PR
+        id: direct-pr
+        if: inputs.pr-number != ''
+        run: |
+          PR_DATA=$(gh pr view ${{ inputs.pr-number }} --json number,headRefName,baseRefName,author --jq '{number: .number, headRefName: .headRefName, baseRefName: .baseRefName, author: .author.login, currentAttempts: 0}')
```
timweri: Same for `${{ inputs.pr-number }}`. This should put it as an env and then use `"$ENV"`.

.github/actions/medic/src/fixer-main.ts
```diff
@@ -0,0 +1,273 @@
...
+    if (!inputs.dryRun) {
+      await incrementAttempts(octokit, owner, repo, inputs.prNumber, inputs.currentAttempts, maxAttempts);
+      try {
+        const { data: pr } = await octokit.rest.pulls.get({ owner, repo, pull_number: inputs.prNumber });
+        await upsertComment(octokit, owner, repo, inputs.prNumber, {
+          success: false,
+          attempt,
+          tokens: claudeResult.tokens,
+          author: pr.user?.login || 'unknown',
+          details: errorMessage,
```
timweri: Error outputs from Claude should be treated caution. For example, it might include phishing markdown injection. I think if Claude fails, the error should be restricted. Overall, Claude open output in comments should not be trusted.

timweri (changes requested):
After looking at this closer, we need more hardenings on the agent:

### Remove access to secrets

We shouldn't let the agent get access to the Github token or any secrets. Use deterministic code to prepare agent inputs and handle agent outputs. The agent step should have no access to secrets

### Sandbox and lock down permissions

We can harden the agent more:

- Run in sandbox mode
- Deny read to sensitive folders
- Limit network access. Maybe with `allowedDomains`.
- Run in `dontAsk` permission mode instead of god mode and allow only needed tools (`Read`, `Edit`)

I think the agent should not be able to call any bash command. It should only read, edit and provide text output. Everything else should be scripted.

PR #1512, opened 2026-04-10
fix(branch-out-upload): add artifact-name input to prevent 409 collision in parallel matrix jobs

Fletch153:
## Summary

- Adds an `artifact-name` input (default: `individual_test_logs`) to the `branch-out-upload` action
- Replaces the hardcoded artifact name with `${{ inputs.artifact-name }}` in the upload step

## Problem

When multiple matrix jobs with `trunk-auto-quarantine: true` fail in the same workflow run, each calls `branch-out-upload` which unconditionally uploads an artifact named `individual_test_logs`. The second job gets a 409 conflict, meaning its diagnostic artifacts are lost.

Observed on `smartcontractkit/chainlink` where `go_core_tests`, `go_core_tests_integration`, and `go_core_ccip_deployment_tests` can collide on the same artifact name.

Note: the upload step only runs under the `failure()` condition, so this does not cause passing jobs to fail — but it does mean we lose test logs from all but the first failing matrix job, making debugging harder.

## Fix

Callers can now pass a unique name per job:
```yaml
artifact-name: ${{ matrix.type.cmd }}_test_logs
```

Backwards compatible — defaults to `individual_test_logs` if not set.

## Related

- Tracked in smartcontractkit/chainlink CORE-2418

### DIFF
--- a/.changeset/brave-donuts-flow.md
+++ b/.changeset/brave-donuts-flow.md
@@ -1,0 +1,5 @@
+---
+"branch-out-upload": patch
+---
+
+Add artifact-name input to prevent 409 collision when multiple matrix jobs upload test logs in the same workflow run
--- a/actions/branch-out-upload/action.yml
+++ b/actions/branch-out-upload/action.yml
@@ -60,6 +60,13 @@ inputs:
       run in different environments (e.g., OSes, browsers, DON topologies).
     required: false
 
+  artifact-name:
+    description: |
+      Name for the uploaded test logs artifact. Override when multiple jobs in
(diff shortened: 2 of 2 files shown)


erikburt (changes requested):
Although this is worth fixing, it is not the cause of failures. The action only uploads when something has already failed.

You must add a changeset ‘pnpm changeset add’ or can try gocs (https://github.com/smartcontractkit/gocs).

jmank88 (dismissed):

erikburt (approved):

PR #1516, opened 2026-04-15
chore: version workflow reusable-docusaurus

erikburt:
This sets up versioning for the `reusable-docusaurus` reusable workflow.

More information on #1448.

### Changes

* Boilerplate for the workflow
* `pnpm workflows:fix`
* Add changeset

---

DX-3229

### DIFF
--- a/workflows/reusable-docusaurus/reusable-docusaurus.yml
+++ b/workflows/reusable-docusaurus/reusable-docusaurus.yml
@@ -1,7 +1,3 @@
-# GENERATED FILE - DO NOT EDIT DIRECTLY.
-# Source: workflows/reusable-docusaurus/reusable-docusaurus.yml
-# Edit the source under workflows/, then regenerate.
-
 name: Reusable Docusaurus
 
 ##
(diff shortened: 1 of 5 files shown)


workflows/reusable-docusaurus/reusable-docusaurus.yml
```diff
@@ -0,0 +1,92 @@
...
+#         build-command: pnpm -C ./docs run build
+#         build-output-directory: ./docs/build
+#         package-json-directory: ./docs
+# ##
+
+on:
+  workflow_call:
+    inputs:
+      # nodejs inputs
+      node-version-file:
```
kalverra, line 29: Does this fail if I don't have a `.tool-versions`?

erikburt: Not sure - this workflow already exists. This versions it so you can reference it with a mutable major version tag. If there's a bug in the workflow then we can update it outside of this PR

kalverra (approved):

PR #1520, opened 2026-04-16
Medic

patrickhuie19:
Addressing all the feedback in https://github.com/smartcontractkit/.github/pull/1447, as well as from a sync with @chainchad and simplifying the initial scope to automatic flakey CI workflow retries.


.github/workflows/medic-workflow-retry-auto.yml
```diff
@@ -0,0 +1,45 @@
...
+      github.event.workflow_run.run_attempt < 3
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Authenticate to GCP
+        uses: google-github-actions/auth@c200f3691d83b41bf9bbd8638997a462592937ed # v2.1.13
+        with:
+          credentials_json: ${{ secrets.GCP_SERVICE_ACCOUNT_KEY }}
```
chainchad, line 35: Can we use federated access with OIDC for GCP instead of hardcoded key similar to how we use AWS IAM roles?

See: workload_identity_provider via https://github.com/google-github-actions/auth

.github/workflows/medic-workflow-retry-auto.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  auto-retry:
+    name: Auto retry failed workflow
+    if: >
+      github.event.workflow_run.conclusion == 'failure' &&
+      github.event.workflow_run.run_attempt < 3
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
chainchad, line 30:
```suggestion
        uses: actions/checkout@v6
```

This solves the node20 issue and for official GH actions, we generally are OK with pinning to a major tag.

.github/CODEOWNERS
```diff
@@ -9,3 +9,8 @@
 
 # data-analytics - temporarily add devex-cicd for dependency bumps
 /actions/pr-quality-check/ @smartcontractkit/data-analytics @smartcontractkit/devex-cicd
+
+# Medic (auto workflow retry only)
+/actions/medic-workflow-retry-auto/** @smartcontractkit/foundations
```
chainchad, line 14: Let's make sure this team has write access to the repo.

.github/CODEOWNERS
```diff
@@ -9,3 +9,8 @@
...
+/libs/medic/** @smartcontractkit/foundations
+/.github/workflows/medic-*.yml @smartcontractkit/foundations
```
chainchad, line 16: We should use a reusable workflow with `.github/workflows/reusable-medic-*` prefix.

.github/medic.yml
```diff
@@ -0,0 +1,18 @@
+# Medic — auto workflow retry (`workflow_run`). Override in consuming repositories.
+# https://github.com/smartcontractkit/.github
+
+# Used for PR author allowlist when deciding whether to auto-retry a failed workflow.
+merge_conflict:
+  allowed_authors:
+    - patrickhuie19
```
chainchad, line 7: Can we specify teams instead of individuals?

actions/medic-workflow-retry-auto/src/main.ts
```diff
@@ -0,0 +1,163 @@
...
+    return { eligible: false, reason: `author ${author} not on allowlist` };
+  }
+
+  return { eligible: true, author, headSha: pr.head.sha };
+}
+
+export async function run(): Promise<void> {
+  try {
+    const token = core.getInput("github-token", { required: true });
+    const octokit = github.getOctokit(token);
```
chainchad, line 60:
```suggestion
    const token = core.getInput("github-token", { required: true });
    core.setSecret(token);
    const octokit = github.getOctokit(token);
```

See: https://github.com/actions/toolkit/blob/main/packages/core/README.md#setting-a-secret

libs/medic/src/workflow-config.ts
```diff
@@ -0,0 +1,124 @@
...
+    return DEFAULT_MEDIC_CONFIG;
+  }
+}
+
+export function parseConfig(content: string): WorkflowRetryConfig {
+  return parseMedicConfig(content).workflow_retry;
+}
+
+export function loadMedicConfigFromFile(
+  filePath: string = ".github/medic.yml",
```
chainchad, line 100: If this is run as a reusable workflow, will other repos have this file within their path? If not, we'd have to clone this repo (.github) within our reusable workflow to make it available.

chainchad: Or we do expect each repo to create this file and this was just added as an example or to be used on this repo (`/.githbub`) directly?

libs/medic/src/retry/claude-client.ts
```diff
@@ -0,0 +1,350 @@
...
+  };
+}
+
+export async function analyzeFailure(
+  octokit: OctokitClient,
+  owner: string,
+  repo: string,
+  runId: number,
+): Promise<AnalysisResult | null> {
+  const projectId = process.env.ANTHROPIC_VERTEX_PROJECT_ID;
```
chainchad, line 202: Should also `core.setSecet(projectId)`.

libs/medic/src/comments.ts
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  let body = `${RETRY_COMMENT_MARKER}
+<p align="center">
+  <img src="${MEDIC_LOGO_URL}" alt="Medic" width="${MEDIC_LOGO_SIZE}" height="${MEDIC_LOGO_SIZE}" />
+</p>
+
+<h2 align="center">${statusEmoji} Medic: Workflow Retry - ${statusLabel}</h2>
+
+@${params.author}
```
chainchad, line 208: Do we want continual pings to the PR author?

libs/medic/src/retry/claude-client.ts
```diff
@@ -0,0 +1,350 @@
...
+
+import * as core from "@actions/core";
+import type { OctokitClient } from "../types";
+import { getWorkflowSummary } from "./github-service";
+import { DEFAULT_TAIL_BYTES, getLogTail } from "./log-service";
+import type { AnalysisResult } from "./analysis-types";
+
+const MODEL = "claude-sonnet-4@20250514";
+const MAX_TOOL_ROUNDS = 5;
+const ANALYSIS_TIMEOUT_MS = 30_000;
```
chainchad, line 13: Sanity check - is this too conservative of a timeout?

libs/medic/src/retry/claude-client.ts
```diff
@@ -0,0 +1,350 @@
...
+        messages.push({ role: "user", content: toolResults });
+        continue;
+      }
+
+      if (response.stop_reason === "end_turn") {
+        const textBlock = response.content.find(
+          (b): b is TextContentBlock => b.type === "text",
+        );
+        if (textBlock?.text) {
+          const result = parseAnalysisResult(
```
chainchad, line 325: Could there be prompt injection in the logs to introduce anything malicious in the PR comments?

libs/medic/src/retry/log-service.ts
```diff
@@ -0,0 +1,100 @@
...
+export async function getLogTail(
+  octokit: OctokitClient,
+  owner: string,
+  repo: string,
+  jobId: number,
+  bytes: number = DEFAULT_TAIL_BYTES,
+): Promise<string> {
+  const logsUrl = await getLogsUrl(octokit, owner, repo, jobId);
+
+  const headResponse = await fetch(logsUrl, { method: "HEAD" });
```
chainchad, line 18: In case there's an error (network call fail) with `fetch()` could the signed URL's leak into the logs? If so we could wrap this with try/catch and share just the message and ensure we're not leaking the URL.

chainchad (changes requested):
Would be good to get @erikburt and others to check too.

.github/medic.yml
erikburt: The default CI token doesn't have read access to org teams/members. The workaround is what CORA does: https://github.com/smartcontractkit/.github/blob/main/.github/workflows/reusable-codeowners-review-analysis.yml#L62-L68 which uses the org-wide members GATI to get a GH token with that access.

But then you have to juggle the 2 tokens, because the org-level token has access to nothing else.

actions/medic-workflow-retry-auto/src/main.ts
erikburt: This will already be marked as a secret by Github itself, or if obtained through another way like GATI will be marked by the GATI action as a secret.

.github/workflows/medic-workflow-retry-auto.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Authenticate to GCP
+        uses: google-github-actions/auth@c200f3691d83b41bf9bbd8638997a462592937ed # v2.1.13
```
erikburt, line 33: Bump to [v3](https://github.com/google-github-actions/auth/releases/tag/v3.0.0)

```
uses: google-github-actions/auth@7c6bc770dae815cd3e89ee6cdf493a5fab2cc093 # v3.0.0
```

actions/medic-workflow-retry-auto/src/main.ts
chainchad: Makes sense. I wonder if we should add it as a safeguard but sounds like it's not critical.

libs/medic/src/workflow-config.ts
```diff
@@ -0,0 +1,124 @@
...
+): MedicConfig {
+  try {
+    const absolutePath = path.resolve(filePath);
+    if (!fs.existsSync(absolutePath)) {
+      core.warning(`Config file not found at ${absolutePath}, using defaults`);
+      return DEFAULT_MEDIC_CONFIG;
+    }
```
erikburt, line 107: Should it default the config? I think having the medic config should be one of the mandatory onboarding steps, and if it doesn't exist, or the existing one is malformed, then perhaps the action should fail.

actions/medic-workflow-retry-auto/src/main.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (workflowRun.conclusion !== "failure") {
+      core.info(
+        `Workflow concluded with "${workflowRun.conclusion}", nothing to retry`,
+      );
+      return;
+    }
+
+    const config = loadMedicConfigFromFile();
+    const retryConfig = config.workflow_retry;
+    const allowedAuthors = config.merge_conflict.allowed_authors;
```
erikburt, line 86: This part of the config seems weird to me because `allowedAuthors` is getting pulled from `merge_conflict` but it's relating to allowed authors for workflow retries.

I would put allowed authors as a top-level field if they're going to be shared across medic's functions.

.github/workflows/medic-workflow-retry-auto.yml
```diff
@@ -0,0 +1,45 @@
...
+concurrency:
+  group: medic-auto-retry-${{ github.event.workflow_run.id }}
+  cancel-in-progress: false
+
+jobs:
+  auto-retry:
+    name: Auto retry failed workflow
+    if: >
+      github.event.workflow_run.conclusion == 'failure' &&
+      github.event.workflow_run.run_attempt < 3
```
erikburt, line 26: This is hardcoded to 3 but the medic config allows for configuring max attempts.

.github/workflows/medic-workflow-retry-auto.yml
chainchad: Also, would this be an org-wide secret? That's the only way this would work as-is. Otherwise, for each repo that we want to use this workflow with, we'd have to set the repo-level secret. If we used OIDC, we could perhaps set it to allow any repo in the org `smartcontractkit` to auth.

actions/medic-workflow-retry-auto/src/main.ts
```diff
@@ -0,0 +1,163 @@
...
+          id: workflowRun.id,
+          name: workflowRun.name,
+          workflowFile,
+          attempt: runAttempt,
+          conclusion: workflowRun.conclusion,
+        },
+        retryConfig.max_attempts,
+      );
+
+      const results: RetryResult[] = [result];
```
erikburt, line 143: Is this created only for the object literal shorthand below? If so, I personally prefer inlining it: `results: [result]`

libs/medic/src/retry/claude-client.ts
```diff
@@ -0,0 +1,350 @@
...
+        () =>
+          reject(
+            new Error(
+              `Analysis timed out after ${ANALYSIS_TIMEOUT_MS / 1000}s`,
+            ),
+          ),
+        ANALYSIS_TIMEOUT_MS,
+      );
+    });
+
```
erikburt, line 250: Loosely related to Chad's comment above.

This is never reset, so it's 30 seconds across all 5 rounds, not 30 seconds per round. Is this intended? It does seem pretty short, especially with GH API calls in between.

libs/medic/src/retry/log-service.ts
```diff
@@ -0,0 +1,100 @@
+/**
+ * Fetches job log tails via signed URL + Range requests.
+ */
+
+import type { OctokitClient } from "../types";
+
+const DEFAULT_TAIL_BYTES = 50_000;
```
erikburt, line 7: `export const DEFAULT_TAIL_BYTES = 50_000;`

erikburt (commented):
Much cleaner than before. Thank you 🙏

PR #1521, opened 2026-04-16
[DX-3787] export slack thread_ts in the slack-notify-git-ref action

Tofel:


chainchad (changes requested):
Make sure to add the changeset file. See https://github.com/smartcontractkit/gocs if you want to avoid `@changesets/cli`.

Tofel:
> Make sure to add the changeset file. See https://github.com/smartcontractkit/gocs if you want to avoid `@changesets/cli`.

@chainchad any chance to add a workflow that would check if changeset have been added? otherwise it is hard to guess it is missing when CI is 🍏

chainchad (approved):

PR #1559, opened 2026-05-19
Add changesets check action for missing changesets

chainchad:
- **Create initial action for Changesets checking**
- **Run formatting**
- **Rebuild**
- **Create changeset**
- **Add pagination support for comments**


actions/cicd-changesets-check/src/run.ts
```diff
@@ -0,0 +1,85 @@
...
+      const changesetFilename = randomBytes(8).toString("hex");
+      const changesetTemplate = [
+        "---",
+        `"${packageName ?? "<package-name>"}": patch`,
+        "---",
+        "",
+        "Description of the change",
+        "",
+      ].join("\n");
+      addChangesetUrl = `${pullRequest.head.repo.html_url}/new/${pullRequest.head.ref}?filename=.changeset/${changesetFilename}.md&value=${encodeURIComponent(changesetTemplate)}`;
```
erikburt, line 48: So this only works on single-package repos? As the root `package.json` may not be the thing we want to bump (like in this repository).

erikburt (approved):

actions/cicd-changesets-check/src/run.ts
chainchad: Yeah, this is for the link to create a markdown file via the GH UI and is more of a convenience thing and can be opted out altogether with the `show-add-link` input.

PR #1589, opened 2026-07-16
Move script to get refs from PR body for reuse

dhaidashenko:
* https://github.com/smartcontractkit/chainlink-evm/pull/504
* https://github.com/smartcontractkit/chainlink-common/pull/2255


erikburt (commented):
How is this intended to be used? Should it be made into an action?

dhaidashenko:
> How is this intended to be used? Should it be made into an action?

The script can be imported into another action to get access to refs. Chainlink-evm and chainlink-common will use it in their respective actions to check whether the Chainlink build breaks on the current commit.
 
I've followed the pattern used with another script [here](https://github.com/smartcontractkit/native-oracle-contracts/blob/44fddaee2dde319d7f661537b66d7cfbce59a402/.github/workflows/solidity-foundry.yml#L363).
Let me know if actions are prefered way of reuse

erikburt:
> The script can be imported into another action to get access to refs. Chainlink-evm and chainlink-common will use it in their respective actions to check whether the Chainlink build breaks on the current commit.
> 
> I've followed the pattern used with another script [here](https://github.com/smartcontractkit/native-oracle-contracts/blob/44fddaee2dde319d7f661537b66d7cfbce59a402/.github/workflows/solidity-foundry.yml#L363). Let me know if actions are prefered way of reuse

Yes, reusable actions are the preferred way. We can merge this now to unblock you though.

erikburt (approved):

erikburt:
Created DX-4670 to track the work on moving this to reusable GHAs instead of clone+invoke style usage.

PR #1596, opened 2026-07-20
Enable Sharing Build Caches + Streamline lefthook

kalverra:
* Enables go build cache sharing in E2E test flows
* Use `stage_fixed` in lefthook to auto-stage changes from pre-commit checks.


tools/scripts/sync-reusable-workflows.sh
chainchad, line 1: Curious why we need to remove this?

kalverra: Because we're now [auto-staging any fix changes with lefthook](https://github.com/smartcontractkit/.github/pull/1596/changes#diff-ad6a01e589b8b1b214ca310dbb8d2e4314f6c612b921050c73c97455de43884dR15), so failing on fixes made and telling me to stage them is now redundant. We still fail on `check` mode.

chainchad (approved):

PR #1608, opened 2026-07-24
[DX-4553] Add docker build cache dance

kalverra:
## Fix `--mount=type=cache`

Use the [buildkit-cache-dance action](https://github.com/reproducible-containers/buildkit-cache-dance) to properly enable caching docker layers with `--mount=type=cache` in CI. See [Docker docs on caching](https://docs.docker.com/build/ci/github-actions/cache/#cache-mounts) for more info.

### 1. Unified `cache-mode` Enum Preset

- Consolidated `docker-restore-cache` and `docker-save-cache` into a single `cache-mode` input parameter with standard presets:
  - `auto` (Default): Save on `push`/`schedule`, restore on `pull_request` and other events.
  - `read-write` (or `true`): Force save and restore.
  - `read-only`: Restore cache only.
  - `write-only`: Save cache only.
  - `off` (or `false`): Disable GHA layer caching.

### 2. Auto-Inferred `buildkit-cache-dance` via `cache-map`

- Callers no longer need to specify `cache-dance: "true"` explicitly. Providing a non-empty `cache-map` automatically enables `buildkit-cache-dance`.
- `cache-map` added as the standard input name, with `cache-dance-cache-map` retained as a backward-compatible fallback alias.


actions/ctf-build-image/action.yml
```diff
@@ -184,7 +216,7 @@ runs:
 
     - name: Build push docker image
       id: build-image
-      uses: smartcontractkit/.github/actions/build-push-docker@build-push-docker/v1
+      uses: smartcontractkit/.github/actions/build-push-docker@7e35079241f460640626adb0ca320fb29aabd8a9 # DEBUG: Using cache dance version
```
erikburt: Should leave this as `/v1`.
* If in the same PR (like right now) there might be a small point at which this is incompatible.
* Or you have to change this action in a separate PR after the above has been retagged to v1.

actions/ctf-build-image/scripts/test-cache-event-overrides.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
```
erikburt, line 1: This script seems like it's testing itself?

actions/build-push-docker/action.yml
```diff
@@ -361,6 +393,15 @@ runs:
...
 
+    # Persist BuildKit RUN --mount=type=cache directories across GHA runs
+    # See: https://github.com/reproducible-containers/buildkit-cache-dance
+    # See: https://docs.docker.com/build/ci/github-actions/cache/#cache-mounts
+    - name: Cache Dance
+      if: ${{ inputs.cache-dance == 'true' }}
+      uses: reproducible-containers/buildkit-cache-dance@5422eac04292c961a382e0f584ea0f03ad9da723 # v3.4.0
+      with:
+        cache-map: ${{ inputs.cache-dance-cache-map }}
+
```
erikburt: Do we need/want to expose this here, or can we just put all of this logic into `ctf-build-image`. Are there plans to use this for other image build workflows that don't get called from `ctf-build-image`?

actions/build-push-docker/scripts/test-cache-dance-config.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
erikburt, line 1: This script also seems to be testing itself.

actions/build-push-docker/action.yml
kalverra: I don't have any immediate plans to use it elsewhere, but I think it makes sense to keep it here for [anyone else who uses this action](https://github.com/search?q=org%3Asmartcontractkit%20smartcontractkit%2F.github%2Factions%2Fbuild-push-docker&type=code) and wants to benefit.

actions/build-push-docker/scripts/test-cache-dance-config.sh
kalverra: Yep, holdover from earlier attempt

actions/build-push-docker/action.yml
kalverra: I don't have any immediate plans to use it elsewhere, but I think it makes sense to keep it here for [anyone else who uses this action](https://github.com/search?q=org%3Asmartcontractkit%20smartcontractkit%2F.github%2Factions%2Fbuild-push-docker&type=code) and wants to benefit.

actions/ctf-build-image/scripts/test-cache-event-overrides.sh
kalverra: Yep, holdover from earlier attempt

actions/build-push-docker/action.yml
```diff
@@ -123,21 +123,29 @@ inputs:
...
+  cache-dance:
+    description: |
+      Deprecated: Explicitly enable/disable buildkit-cache-dance. If unset, cache-dance is automatically enabled when cache-map is provided.
+    required: false
+    default: ""
+  cache-dance-cache-map:
+    description: |
+      Deprecated alias for cache-map.
+    required: false
+    default: ""
```
erikburt: These are "deprecated" but they were never released in the first place.

erikburt (commented):
Would like to see the `/chainlink` pr tested against the new changes too.

actions/ctf-build-image/scripts/go-get-overrides.sh
```diff
@@ -0,0 +1,45 @@
...
+  dep="${line%%=*}"
+  sha="${line#*=}"
+  [[ -z "$dep" || -z "$sha" ]] && continue
+
+  # Qualify module name if short name provided
+  if [[ "$dep" != *"/"* ]]; then
+    module="github.com/smartcontractkit/${dep}"
+  else
+    module="${dep}"
+  fi
```
erikburt, line 29: Is there a use-case for supporting arbitrary overrides, if not I think limiting it to smartcontractkit deps makes the most sense to me.

kalverra: > Is there a use-case for supporting arbitrary overrides

Probably not, but will ask @jmank88 to check.

actions/build-push-docker/action.yml
```diff
@@ -359,12 +350,24 @@ runs:
...
+      id: resolve-cache
       shell: bash
       env:
+        ACTION_PATH: ${{ github.action_path }}
+        CACHE_MODE: ${{ inputs.cache-mode }}
+        GITHUB_EVENT_NAME: ${{ github.event_name }}
         DOCKER_RESTORE_CACHE: ${{ inputs.docker-restore-cache }}
         DOCKER_SAVE_CACHE: ${{ inputs.docker-save-cache }}
+      run: |
+        "${ACTION_PATH}/scripts/resolve-cache-settings.sh" >> "$GITHUB_OUTPUT"
```
erikburt: I would prefer `| tee -a "${GITHUB_OUTPUT}"`. I find it very helpful for debugging.

actions/build-push-docker/action.yml
```diff
@@ -200,9 +190,10 @@ runs:
       shell: bash
       env:
         ACTION_PATH: ${{ github.action_path }}
-        CACHE_DANCE: ${{ inputs.cache-dance }}
-        CACHE_MAP: ${{ inputs.cache-map || inputs.cache-dance-cache-map }}
+        CACHE_MAP: ${{ inputs.cache-map }}
+        CACHE_DANCE_CACHE_MAP: ${{ inputs.cache-dance-cache-map }}
       run: |
+
         "${ACTION_PATH}/scripts/validate-cache-dance-config.sh" >> "$GITHUB_OUTPUT"
```
erikburt: `| tee` here as well.

actions/build-push-docker/action.yml
```diff
@@ -136,14 +131,9 @@ inputs:
...
-    description: |
-      Deprecated: Explicitly enable/disable buildkit-cache-dance. If unset, cache-dance is automatically enabled when cache-map is provided.
-    required: false
-    default: ""
   cache-dance-cache-map:
     description: |
-      Deprecated alias for cache-map.
+      [Deprecated] Alias for `cache-map`. JSON string mapping cache mount paths/names for buildkit-cache-dance.
     required: false
     default: ""
```
erikburt: Can't this be outright deleted? It was never released, we don't need to deprecate an input before it's actually released/used.

actions/build-push-docker/README.md
```diff
@@ -30,12 +30,8 @@ multi-platform manifest from per-arch builds.
...
-| `docker-build-cache-to`   | no           | GHA cache scoped to OS/arch                           | Override the cache destination. Effective only when `docker-save-cache` is `true`.                                                                                                                                                                                          |
+| `docker-build-cache-from` | no           | GHA cache scoped to OS/arch                           | **DEPRECATED: Use `cache-mode` instead!** Override the cache source. Effective only when cache restore is enabled via `cache-mode` (e.g. `auto` on PRs, `read-write`, or `read-only`).                                                                                      |
+| `docker-build-cache-to`   | no           | GHA cache scoped to OS/arch                           | **DEPRECATED: Use `cache-mode` instead!** Override the cache destination. Effective only when cache save is enabled via `cache-mode` (e.g. `auto` on push/schedule, `read-write`, or `write-only`).                                                                         |
```
erikburt: These aren't deprecated.

actions/ctf-build-image/action.yml
```diff
@@ -123,10 +123,20 @@ inputs:
...
+    required: false
+    description: |
+      [Deprecated] Toggles docker cache restore for builds. Use `cache-mode` instead.
+    default: ""
+  docker-save-cache:
+    required: false
+    description: |
+      [Deprecated] Toggles docker cache save after builds. Use `cache-mode` instead.
+    default: ""
   cache-mode:
```
erikburt: Why add deprecated inputs to this action?

actions/build-push-docker/action.yml
erikburt: If it's interfering with exit codes, then the script can call tee directly

actions/build-push-docker/scripts/validate-cache-dance-config.sh
```diff
@@ -0,0 +1,32 @@
...
+    exit 1
+  fi
+  if ! command -v jq >/dev/null 2>&1; then
+    echo "::error::'jq' command not found; required to validate cache-map JSON."
+    exit 1
+  fi
+  if ! echo "$CACHE_MAP" | jq . >/dev/null 2>&1; then
+    echo "::error::cache-map must be valid JSON."
+    exit 1
+  fi
```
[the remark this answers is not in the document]
kalverra: cache-dance action checks this already

erikburt (approved):

PR #1620, opened 2026-08-04
Remove `vulnerability-alerts: read`

kalverra:
Not sure why this was here, but it's not valid.

Also made lefthook a little prettier output.


jmank88 (changes requested):
It is valid, and it's required for the API we are accessing
Edit: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

PR #1626, opened 2026-08-05
workflows/reusable-dependabump: fix GOTOOLCHAIN=auto

jmank88:


chainchad (dismissed):
Don't forget the changeset

chainchad (approved):

PR #1634, opened 2026-08-11
[DX-4969] Handle merge_group events

kalverra:
`merge_group` events had issues with git history, as we didn't fetch the proper git history. The `git pull` fallback [didn't work properly either](https://github.com/smartcontractkit/chainlink/actions/runs/31510996276/job/93844475256#step:3:140). Now we fetch proper history with the API.

Tested on https://github.com/smartcontractkit/chainlink/pull/23386

### DIFF
--- a/actions/advanced-triggers/src/github.ts
+++ b/actions/advanced-triggers/src/github.ts
@@ -42,18 +42,25 @@ export async function getChangedFilesForMergeGroup(
     `Fetching changed files for ${owner}/${repo} merge group ${base}...${head}`,
   );
 
-  const files = await octokit.paginate(
-    octokit.rest.repos.compareCommits,
-    {
-      owner,
-      repo,
-      base,
-      head,
-      per_page: 100,
-    },
-    (response: CompareResponse) =>
-      response.data.files?.map((f: CompareFiles[number]) => f.filename) ?? [],
-  );
+  const res = await octokit.rest.repos.compareCommits({
+    owner,
+    repo,
+    base,
+    head,
+    per_page: 100,
+  });
 
-  return files;
+  if (!res.data.files) {
+    throw new Error(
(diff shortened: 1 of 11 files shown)


actions/advanced-triggers/src/github.ts
```diff
@@ -25,3 +30,30 @@ export async function getChangedFilesForPR(
...
+  repo: string,
+  base: string,
+  head: string,
+): Promise<string[]> {
+  core.debug(
+    `Fetching changed files for ${owner}/${repo} merge group ${base}...${head}`,
+  );
+
+  const files = await octokit.paginate(
+    octokit.rest.repos.compareCommits,
```
chainchad: We're paginating commits here but IIRC there's a limit on `response.data.files` to a hard cap like 300.

actions/advanced-triggers/src/github.ts
```diff
@@ -25,3 +30,30 @@ export async function getChangedFilesForPR(
...
+    {
+      owner,
+      repo,
+      base,
+      head,
+      per_page: 100,
+    },
+    (response: CompareResponse) =>
+      response.data.files?.map((f: CompareFiles[number]) => f.filename) ?? [],
```
chainchad: `response.data.files?.map(...) ?? []` treats an absent files field as "no files changed" but this probably isn't what we want. Should we throw in that case and let the fallback handle it?

actions/advanced-triggers/src/github.ts
chainchad: https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit

chainchad (changes requested):

chainchad (approved):

PR #1638, opened 2026-08-18
Support bun for optional package management on cicd-changesets

chainchad:
Needed for some TypeScript templates that get scaffolded with bun in order to avoid having to mix pnpm with bun.

### DIFF
--- a/actions/cicd-changesets/scripts/resolve-bun-version.sh
+++ b/actions/cicd-changesets/scripts/resolve-bun-version.sh
@@ -116,4 +116,4 @@ fi
 {
   echo "version=${out_version}"
   echo "file=${out_file}"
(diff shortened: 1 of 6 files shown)


actions/cicd-changesets/scripts/resolve-bun-version.sh
```diff
@@ -0,0 +1,119 @@
...
+    else
+      echo "No bun pin found; setup-bun will read package.json or install latest"
+    fi
+  fi
+fi
+
+{
+  echo "version=${out_version}"
+  echo "file=${out_file}"
+} >>"${GITHUB_OUTPUT}"
```
kalverra: Use `| tee -a`

kalverra (approved):

PR #1642, opened 2026-08-24
IP-3478]: Remove Bitnami Chart Refs

tvc-jaz:
## What

Remove the Bitnami Helm chart repo from the shared `ctf-setup-run-tests-environment` action 

## Why

No chart dependency in scope requires the Bitnami repo — the action never installs a
chart itself, and the `chainlink-cluster` chart (from `chainlink-qa`) has no Bitnami
dependency. This aligns with the org-wide Bitnami removal effort and is the ticket's
preferred approach (remove when unused).

## Changes

- **actions/ctf-setup-run-tests-environment/action.yml** — drop the `helm repo add
  bitnami` line; `chainlink-qa` repo add unchanged.
- **.changeset/ctf-setup-remove-bitnami-repo.md** — add `minor` changeset.

### DIFF
--- a/.changeset/ctf-setup-remove-bitnami-repo.md
+++ b/.changeset/ctf-setup-remove-bitnami-repo.md
@@ -1,0 +1,5 @@
+---
+"ctf-setup-run-tests-environment": minor
+---
+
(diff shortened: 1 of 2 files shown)


chainchad (approved):
LGTM but would also like @skudasov to review

skudasov (approved):

PR #1656, opened 2026-09-08
fix(docker): don't fail-fast the build-publish arch matrix

HashWrangler:
## Problem

`build-publish` runs one job per arch under a matrix with no `fail-fast` key, so it defaults to `true`. Each arch pushes its **own per-arch tag** before `docker-manifest` assembles them. That ordering makes a cancellation destructive:

1. arm64 fails (in our case a transient `proxy.golang.org` read error mid-build)
2. fail-fast cancels amd64 **one second later** — but amd64 had already completed its build and pushed `<tag>-amd64`
3. `docker-manifest` skips, since it needs both arches
4. Every retry now fails instantly on the `One or more provided tags already exist in ECR` pre-check, because `<tag>-amd64` is there

The result is an image that is permanently half-published: the per-arch tag exists, the manifest tag never does, and no re-run can recover it. The only ways out are deleting the orphaned tag from ECR or abandoning the version.

Real occurrence today on `public.ecr.aws/chainlink/ccip:2.64.0-ccip-beta.0` — `-amd64` present, `-arm64` and the manifest absent, three re-run attempts all failing on the guard.

## Change

`fail-fast: false` on that matrix. A transient failure in one arch now leaves the other `success` instead of `cancelled`, so `gh run rerun --failed` retries only the broken arch and the manifest step can finish.

## Blast radius

Failure handling only, in the less-aggressive direction. Worst case a doomed build burns a few extra runner-minutes on the sibling arch instead of being cancelled early. No change to inputs, outputs, tagging, or publish behaviour.

Applied to both the consumable copy and the package source. Validated with `actionlint`; both files parse.


HashWrangler:
Should be an easy one @chainchad .  Fixes a flake state where a fail on ARM can cause an orphaned tag when AMD succeeds.  This guard would allow a rerun of just the failed architecture rather than orphaning a good build on the other.

HashWrangler:
> ### 🟢 Approval recommended
> The change is minimal, correctly scoped to matrix failure behavior, and aligns with the described operational failure mode without altering build inputs/outputs.
> 
> Pull request overview
> This PR prevents destructive cancellations in the `build-publish` multi-arch Docker workflow by explicitly disabling matrix fail-fast, ensuring that a transient failure in one architecture doesn’t cancel the sibling build after it has already pushed its per-arch tag (which previously could leave images half-published and non-rerunnable).
> 
> **Changes:**
> 
> * Set `strategy.fail-fast: false` for the `build-publish` arch matrix in the reusable workflow definition.
> * Apply the same change to both the package source workflow and the consumable `.github/workflows` copy.
> * Add a changeset documenting the motivation and operational impact.
> 
> File summaries
> File	Description
> workflows/reusable-docker-build-publish/reusable-docker-build-publish.yml	Disable matrix fail-fast for `build-publish` to avoid cancelling sibling arch builds mid-publish.
> .github/workflows/reusable-docker-build-publish.yml	Mirror the same `fail-fast: false` change in the consumable workflow copy.
> .changeset/docker-build-publish-no-fail-fast.md	Document the behavior change and why it prevents half-published images.
> Review details
> * **Files reviewed:** 3/3 changed files
> * **Comments generated:** 0
> * **Review effort level:** Lite
> 
> 💡 [Add a `code-review` agent skill](/smartcontractkit/.github/new/main?filename=.github/skills/code-review/SKILL.md) or configure MCP servers for context-aware, tailored reviews. [Learn more in the docs.](https://docs.github.com/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review?tool=webui#mcp-servers-and-agent-skills)

Aww. Thanks copilot.  That's the nicest thing you've ever said to me.

chainchad (approved):
👍 👍

